Issue #1, opened 2020-10-23
Exemplary issue


lukasz-starosta:
## Description
Hej, gdzie jest pepe?

## Environment
> Specify your system, browser and tool versions if necessary

Windows 10, Chrome 86, Node v10.16

## Correct issue checklist:
- [x] I have added a detailed description
- [x] I have specified the environment
- [ ] I have assigned a Team label (front-end / node / git / design / gatsby)
- [x] I have assigned a Description label (enhancement / bug / etc.)
- [ ] I have assigned a Project

Issue #3, opened 2020-11-02
Publication configuration file


karczewskiPiotr:
## Description
I'd like to start a discussion on how the configuration file for a web publication should look like.
### Content
Things that are for sure relevant:
* Full name of the publication
* Collaborators with roles:
    * programmer
    * editor
* Package manager choice

Maybe it would be a good practice to dump all the preferences, which user chose to create the project, into this file for any future reference on how it was set up? Those could include:
* Any details for `gatsby-config` specified during set up
* Chosen technology extensions

### Format
Should we go for JSON or YAML? Doesn't make much of a difference in my opinion but it's something to choose.

#### JSON
```json
{
  "project_name": "Example project name",
  "collaborators": [
    {
      "github_username": "wuja",
      "role": "programmer"
    },
    {
      "github_username": "john",
      "role": "editor"
    }
  ],
  "pm_preference": "yarn"
}
```
#### YAML
```yaml
project_name: Example project name
collaborators:
  - github_username: wuja
    role: programmer
  - github_username: john
    role: editor
pm_preference: yarn

```

## Correct issue checklist:
- [x] I have added a detailed description
- [ ] I have specified the environment
- [x] I have assigned a Team label (front-end / node / git / design / gatsby)
- [x] I have assigned a Description label (enhancement / bug / etc.)
- [x] I have assigned a Project
- [ ] I have added a Milestone (if applicable)

karczewskiPiotr:
In todays meeting, the issue of multi-edition publication was raised. So basically what if I have multiple publications falling under the same collection?

It would be nice to group them in the projects list. In order to achieve that we can simply tag the project and leave the trace in the configuration file so that it becomes an attribute of the project state on the front-end where projects can be easily grouped.

Therefore a `tag` field becomes another necessary field in the configuration file.

lukasz-starosta:
I suggest here to go for JSON:
* easier to transfer by text
* will not be messed up because of whitespace
* easier to parse in the browser, without external libraries  

We might also want to use a JSON Schema to keep the format of our config files consistent.

PR #14, opened 2020-11-17
Creation of basic fileSystem functions and recursive publication search

### LINKED ISSUE #13: __Determine which directories on local machine are web publication projects based on the config file presence__

KrzysztofMoszczynski:
## Description
This pull request implements basic file explorer functions and recursive search of publication projects. 

Small note: There is a function checking if a directory is a github repository. It's there only for optimization. The algorithm checks if current dir has the publication_config file, if not, it checks if it is a github repository. If so, the function returns not to search the publication_config file in plenty of folders which it won't be found for sure.
## Linked issues
closes #13 

## Correct PR Checklist
- [ ] I have assigned Reviewers
- [ ] I have assigned a Team label
- [ ] I have assigned an Description label
- [ ] I have assigned a Project

### DIFF
--- a/lib/file-manager.ts
+++ b/lib/file-manager.ts
@@ -1,0 +1,95 @@
+var fs = require('fs');
+var path = require('path');
+
+var currentPath: string = "";
+
+function getCurrentPath(): string {
+    return currentPath;
+}
+
+function setCurrentPath(newPath: string): void {
+    currentPath = newPath;
+}
+
+function enterFolder(newFolder: string): void {
+    var newPath: string = path.join(getCurrentPath(), newFolder);
+    setCurrentPath(newPath);
+}
+
+function escapeFolder(currentFolder: string): string {
+    var newPath = path.dirname(currentFolder);
+    return newPath;
+}
+
+function escapeCurrentFolder(): void {
+    var newPath: string = path.dirname(currentPath);
+    setCurrentPath(newPath);
+}
+
+function checkIfRepository(source: string): boolean {
+    if(fs.existsSync(path.join(source, ".git"))) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+function checkIfPublication(source: string): boolean {
+    if(fs.existsSync(path.join(source, "publication_config"))) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+function checkIfDirectory(source: string): boolean {
+    try {
+        if(fs.lstatSync(source).isDirectory()) {
+            return true;
+        } else {
+            return false;
+        }
+    } catch(error) {
+        console.log(error);
+        return false;
+    }
+}
+
+function getDirectories(source: string): string[] {
+    var directories: string[] = [];
+    try {
+        const isDirectory = source => fs.lstatSync(source).isDirectory()
+        directories = fs.readdirSync(source).map(name => path.join(source, name)).filter(isDirectory);
+    } catch(error) {
+        console.log(error);
+    }
+    return directories;
+}
+
+function findAllPublications(path: string): string[] {
+    var listOfProjects: string[] = [];
+    if(fs.existsSync(path)) {
+        if(checkIfDirectory(path)) {
+        recursiveSearch(path, listOfProjects);
+        }
+    } else {
+        console.log("There is no such directory or file!");
+    }
+    return listOfProjects;
+}
+
+function recursiveSearch(path: string, listOfProjects: string[]): void {
+    if(checkIfPublication(path)) {
+        listOfProjects.push(path);
+        return;
+    } else if(!checkIfRepository(path)) {
+        const availableDirectories: string[] = getDirectories(path);
+        for(var i = 0; i < availableDirectories.length; i++) {
+            recursiveSearch(availableDirectories[i], listOfProjects);
+        }
+    }
+    return;
+}
+
+module.exports = {findAllPublications};
+


lib/file-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+    } else if(!checkIfRepository(path)) {
+        const availableDirectories: string[] = getDirectories(path);
+        for(var i = 0; i < availableDirectories.length; i++) {
+            recursiveSearch(availableDirectories[i], listOfProjects);
+        }
+    }
+    return;
+}
+
+module.exports = {findAllPublications};
```
karczewskiPiotr, line 94: Could have just had `export function findAllPublications() {...}` .

karczewskiPiotr (changes requested):
Remember to not log stuff to the console, either implement custom errors or just do what suits the purpose like return empty array. 

The `findAllPublications` should probably be `findLocalPublications` as it doesn't give back the ones remotely on GitHub and if you have boolean returning methods or variables maybe it's good to stick to convention and do `isDirectory`, `isRepository`, `isPublication`.

It would be good if the `findAllPublication` function returned actual publications basing on the config file instead of just strings with path. Try to define a Publication type based on the discussion in #3 and return an array of the ones composed from contents of publication config file. Could also already try and implement a simple slice for storing the result in state.

When it comes to using Redux here I will provide and example based on this code so we can go over it on a meeting of some sorts.

pkazmierski (changes requested):
That's something to discuss with @karczewskiPiotr, but shouldn't the exported function be `async`/return a `Promise`? Perhaps this is somehow handled already because of the main/renderer split, but normally running a synchronous function will stall the main (UI) thread.

It's a good practice to use `let ` instead of `var`, unless you specifically need `var`.

Missing jsdoc - it should be added at least to the exported function.

Don't `path` and `fs` packages have their TS types available? If so, please add them using `yarn add @types/<package_name>`.

lib/file-manager.ts
```diff
@@ -0,0 +1,95 @@
+var fs = require('fs');
+var path = require('path');
```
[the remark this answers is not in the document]
KrzysztofMoszczynski: There are problems with using import in node. The basic syntax is require and in works everywhere. I have read that it is needed to have --experimental-modules flag to use it and I just don't know if it's worth it.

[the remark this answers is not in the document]
KrzysztofMoszczynski: ok

karczewskiPiotr:
Anything more happening here @KrzysztofMoszczynski? I can see some new commits but the PR is still set to *work in progress* and has conflicts with main branch. 🕵️

src/renderer/components/description/description.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const currentUser = useSelector(selectCurrentUser);
+  const path = useSelector(selectDefaultDirPath);
+
+  // log user to showcase state changes if you don't have Redux Tools
+  console.log(currentUser);
+  console.log(path);
+  console.log(findLocalPublications());
+  const gatsbyInstallStuff = useSelector(selectGatsbyInstallStatus);
+
+  // log user to showcase state changes if you don't have Redux Tools
```
lukasz-starosta, line 29: why is this duplicated?

src/shared/slices/publicationsSlice.ts
```diff
@@ -0,0 +1,34 @@
...
+    dirPath: string, 
+    publicationName: string
+    };
+
+type PublicationList = {
+        list: [PublicationListElement]
+    };
+
+
+  const initialState: PublicationList = {list:[{id: 0, dirPath:"", publicationName: ""}]};
```
lukasz-starosta, line 15: why is this PublicationList an object?

src/renderer/components/description/description.tsx
```diff
@@ -0,0 +1,59 @@
...
+import './description.scss';
+import { useDispatch, useSelector } from 'react-redux';
+import {
+  selectCurrentUser,
+} from '../../../shared/slices/currentUserSlice';
+import {
+  selectDefaultDirPath,
+  setDefaultDirPath,
+} from '../../../shared/slices/configurationSlice';
+import findLocalPublications from '../../../main/node/file-manager';
```
lukasz-starosta, line 11: you do `export default {findLocalPublications, isDirectory, isPublication}` but import this whole object as `findLocalPublications` here? are you sure?

src/renderer/components/description/description.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+// TODO: delete exemplary buttons from here
+const Description = () => {
+  const currentUser = useSelector(selectCurrentUser);
+  const path = useSelector(selectDefaultDirPath);
+
+  // log user to showcase state changes if you don't have Redux Tools
+  console.log(currentUser);
+  console.log(path);
+  console.log(findLocalPublications());
```
lukasz-starosta, line 26: what is this supposed to do?

src/renderer/components/description/description.tsx
lukasz-starosta: https://github.com/karczewskiPiotr/vnlab-tool/pull/14/commits/d88c5b71a899d8138b92c372c455365d5354e667
I think this commit changes affect this. Did you test this code after this change?

karczewskiPiotr (changes requested):
Seems like merge with main failed completely. There are duplicate files here that no longer exist in the project. 

Firstly, exports and imports are mismatched like @lukasz-starosta mentioned. The `findLocalPublications` function can't be called because of this (and shows error even in VS Code). Even when it's called after fix of the import it does not work at all and throws error instantly because of use of hook outside a function. Haven't gone any deeper so not sure if that is where the problems would end, can't really check anything if it doesn't work.

Have you tested this code? Why even request a review if it doesn't work? 🤔 

None of the functionality is also async, which was a concern raised by @pkazmierski before and went over during the meeting. This function shouldn't be called in renderer process as well but there is no async Redux action to call it in main process like we talked about in the meeting the day this PR was opened.

Either way not going to look at specifics of the implementation as it just doesn't work so try to fix it first.

src/renderer/components/description/description.tsx
KrzysztofMoszczynski: I failed with merging my branch with main, because my branch had the descrption file while the main didn't have it, but I don't know how to delete the files nonexistent main files without deleting my changes

src/shared/slices/publicationsSlice.ts
KrzysztofMoszczynski: I did't know how to export the list without making it an object, so I could export the "list" field.

src/renderer/components/description/description.tsx
KrzysztofMoszczynski: same as above, the description file shouldn't be here. I'm just gonna delete it

src/renderer/components/description/description.tsx
KrzysztofMoszczynski: same

src/renderer/components/description/description.tsx
lukasz-starosta: Ok, just delete the duplicate

msuliborski (approved):

pkazmierski (approved):

src/main/node/file-manager.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+/**
+ * Finds all publication config files and uploads their content to the application state.
+ * Rejects the promise if a command execution error occurs.
+ * @param {string} source
+ * @return {Promise<boolean>}
+ */
+export function findLocalPublications(source: string): Promise<boolean> {
```
lukasz-starosta, line 73: This is not asynchronous in any way right? Does it affect the application / stop it for a significant amount of time?

KrzysztofMoszczynski: I thought I can just use `return new Promise()` to make it asynchronous. Do I have to make it "async, await"?

pkazmierski: @lukasz-starosta It searches for publications recursively in a folder structure. All file access operations must be async, since it can take a lot of time, either because of the number of folders to search or having slow data access (network storage over SFTP, HDDs etc.).

lukasz-starosta (approved):

PR #16, opened 2020-11-17
Function for installing gatsby-cli via NPM

### LINKED ISSUE #15: Allow global installation of gatsby-cli via NPM
Allow installation of gatsby-cli (if necessary) via NPM, which is called in a child process.

pkazmierski:
## Description
Adds a function for installing gatsby-cli via NPM. Also adds @types/node (for proper node.js type support in TS) and child_process (for running child processes) packages to yarn.

## Linked issues
Closes #15

## Correct PR Checklist

### DIFF
--- a/lib/gatsby.ts
+++ b/lib/gatsby.ts
@@ -1,5 +1,10 @@
 import { exec } from 'child_process';
 
+
+/**
+ * Check if gatsby-cli is installed globally.
+ * @return {Promise<boolean>}
+ */
 function checkForGatsby(): Promise<boolean> {
   return new Promise<boolean>((resolve, reject) => {
     exec('npm list -g gatsby-cli --json', (error, stdout, stderr) => {
@@ -13,6 +18,10 @@ function checkForGatsby(): Promise<boolean> {
   });
 }
 
+/**
+ * Checks if gatsby is already installed globally and if not installs it.
+ * @return {Promise<void>}
+ */
 export function installGatsby(): Promise<void> {
   return new Promise<void>((resolve, reject) => {
     checkForGatsby().then(
--- a/lib/gatsby.ts
+++ b/lib/gatsby.ts
@@ -1,5 +1,14 @@
 import { exec } from 'child_process';
+import * as winston from 'winston';
 
+const myFormat = winston.format.printf(({ level, message, label, timestamp }) => {
+  return `${timestamp} ${level}: ${message}`;
+});
+
+const logger = winston.createLogger({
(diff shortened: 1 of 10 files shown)


lib/gatsby.ts
```diff
@@ -0,0 +1,47 @@
...
+          console.info('gatsby-cli is already installed');
+          resolve();
+        } else {
+          console.info('gatsby-cli not installed. Installing...');
+          exec('npm install -g gatsby-cli --quiet', (error, stdout, stderr) => {
+            if (error) reject(error);
+
+            console.log(`stdout: ${stdout}`);
+            console.error(`stderr: ${stderr}`);
+            console.info('Finished installation of gatsby-cli');
```
karczewskiPiotr, line 39: Unsure if there is any value to logging this stuff in the console here. Any idea how this behaves when the app is built?

On the other hand, it might be a good idea to save logs only if install fails. Dumping it into a file could maybe provide some value if for some reason this fails in production and user want's to find out what went wrong?

What do you think? 🤔

pkazmierski: The way it works now is that you can see all of the output in the dev tools console after the installation finishes. I would keep some logging capabilities because:
- as you have said, if some problems occur it will help with debugging
- gatsby-cli installation is a one-time thing and there isn't a lot of info to log, so the performance and log size cost are negligible
That being said, logging to a file is better, so I'll change it.

karczewskiPiotr: If we run this in main thread (either by sending signal via IPC or using `createActionMain` helper to create Redux action) the output will not show up in devtools as it's only run in main process and it would output to that console. So the question would be if it opens up that console in production.

We might also want to later attach this function to some state that front-end team could use to either show that it's installing but I'll consult with them about this.

karczewskiPiotr (approved):

src/renderer/components/description/description.tsx
```diff
@@ -7,13 +7,19 @@ import {
...
+} from '../../../shared/slices/gatsbyInstallSlice';
 
 // TODO: delete exemplary buttons from here
 const Description = () => {
   const currentUser = useSelector(selectCurrentUser);
+  const shit = useSelector(selectGatsbyInstallStatus);
 
   // log user to showcase state changes if you don't have Redux Tools
   console.log(currentUser);
+  console.log(shit);
```
lukasz-starosta, line 22: 🙈

karczewskiPiotr (approved):

PR #19, opened 2020-11-22
Custom navigation

### LINKED ISSUE #5: Figure out how to tackle navigation
Considerations
* If using `react-router` is `MemoryRouter` the best option?
* Are there any valid alternatives to `react-router`?

Eleara:
## Description
Created custom routing and added exemplary components with a basic navigation bar.
To add a view one has to add the component to componentsEnum.ts and to the switch in getRenderedView method in custom_router.tsx.

## Linked issues
closes #5

## Correct PR Checklist


szymonrucinski (dismissed):
Seems fine.

src/renderer/components/app/navigation_bar/navigation_bar.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }
+  
+  const renderListOfButtons = () => {
+    var listOfButtons:JSX.Element[] = [];
+    Object.keys(COMPONENTS_LIST).forEach((key) => {
+      const buttonKey = getKeyValue(key)(COMPONENTS_LIST);
+      listOfButtons.push(<button onClick={onNavigationButtonClick(buttonKey)}>{buttonKey}</button>)
+    })
+    return listOfButtons;
+  }
```
karczewskiPiotr, line 27: Could just `map` over the keys in cases like these instead of `forEach` so that there is no need for any variables.

```tsx
const renderListOfButtons = () => {
    return Object.keys(COMPONENTS_LIST).map((key) => {
      const buttonKey = getKeyValue(key)(COMPONENTS_LIST);
      return (
        <button key={key} onClick={onNavigationButtonClick(buttonKey)}>
          {buttonKey}
        </button>
      );
    });
  };
```

One more thing is that if you iterate over an object or an array to return elements React requires those elements to have a unique `key` (error must have been shown in dev tools) so in this case it can just be the component name as it's unique but in cases where you don't have such key you can use index of the mapped element.

src/shared/slices/currentViewSlice.ts
```diff
@@ -0,0 +1,27 @@
...
+      state.view = action.payload;
+    },
+  },
+});
+
+// export actions from slice
+export const { updateCurrentView } = CurrentViewSlice.actions;
+
+// selector for current user | note the use of RootState type here, it's necessary as selectors access whole state of the store
+export const selectCurrentView = (state: RootState) => state.currentView;
```
karczewskiPiotr, line 25: This could maybe already select the `view` value if nothing else is stored in the state but I suppose it will be expanded for the future. 💁‍♀️

karczewskiPiotr (changes requested):
Why are all the components in the `app` component folder?

The temporary implementation of the authentication by `UserLogin` component shows responsibility problems. In the current state it is responsible for rendering the whole application with the use of `CustomRouter`. Would be nice if this functionality was either extracted to a HOC or some sort of wrapper component like:

```tsx
const Auth: React.FC = ({children}) => { //could type this with props type that has children type for full Typescript goodness
  return (
    // This could check for some state value of currentUser in the future when such slice is implemented but that's Jakub and Adam's job
    return isUserLoggedIn() ? <>{children}</> : <LoginComponent/>;
  )
}

const App = () => {
  return (
    <Provider store={store}>
      <Auth>
        /* render these components if authenticated */
      </Auth>
    </Provider>
  );
};
```

The `CustomRouter` is then also responsible for rendering the `NavigationBar` bar, which really shouldn't be the case as it's a standalone element displayed on the screen all the time and that doesn't rely on any functionality given by `CustomRouter`. I'd say stick to a principle (within reason) of as small components as possible and make sure they are responsible for one thing only. So following in the direction of previous code block it could be done along the lines of this

```tsx
const App = () => {
  return (
    <Provider store={store}>
      <Auth>
        <div className="some-layout-wrapper">
          <NavigationBar />
          <CustomRouter />
        </div>
      </Auth>
    </Provider>
  );
};
```

The implementation of the `CustomRouter` itself also seems not elegant with the switch. Since this system doesn't seem to allow for multiple routers and passing props in any way it can leverage the state held in slice even more to reduce the long switch that would be created in case of many views.

If we restructure the views info which now stores strings for switch condition to check against we could come to something along the lines of:

```ts
impor Description ....

type RouterComponents = {
  [key: string]: {
    view: string;
    component: React.FC;
  };
};

export const components: RouterComponents = {
  description: {
    view: 'description', // this could be specified with enum if desired but needs to match object key
    component: Description,
  },
};
```

which would simplify the router to this I presume:

```tsx
const CustomRouter = () => {
  const currentView = useSelector(selectCurrentView);

  const View: React.FC = components[currentView.view].component;

  return (
    <div className='router'>
      <div className='sideBar'>
        <NavigationBar />
      </div>
      <div className='content'>
        <View />
      </div>
    </div>
  );
};
```
There is definitely room for improvement here. These are suggestions for you guys to think about and implement if you agree that a similar direction would better the solution but for sure this PR should not be merged with this file structure and component responsibility issues.

lukasz-starosta:
Apart from @karczewskiPiotr's remarks, I would like to suggest another convention of project structure.  
First: our components are called `MyComponent` but the files are `my_component` - is there any reason for this? If not, I would rather that these names are the same to avoid confusion.  
Second: move all of the components into proper folders, as right now they are all in `/app`

217853:
I looked through the changes, I agree with the comments and suggestions of Łukasz and Piotrek, structure is a little derailed. All components are in one place /app which may be intuitive for now, but when the components begin to expand their code, it may be hard to search through. Better to create some separate folders like Łukasz suggested. As for pure code, I didnt notice anything else that wasn't mentioned above.

217853 (commented):
I looked through the changes, I agree with the comments and suggestions of Łukasz and Piotrek, structure is a little derailed. All components are in one place /app which may be intuitive for now, but when the components begin to expand their code, it may be hard to search through. Better to create some separate folders like Łukasz suggested. As for pure code, I didnt notice anything else that wasn't mentioned above.

lukasz-starosta (changes requested):
As mentioned above, sorry for duplicate comments

lukasz-starosta (dismissed):
Looks fine to me, let's close it so we can move on.  
I think we should do as much of the project as possible until the deadline and leave some minor code style fixes for later.

217853 (dismissed):

karczewskiPiotr (approved):

Issue #23, opened 2020-11-25
__Read remote GitHub projects that are web publications__


aleksanderbrylski:
Perform search based on the presence of configuration file.

PR #25, opened 2020-11-25
Git/authenticate user

### LINKED ISSUE #20: Authenticate user from GitHub account

aleksanderbrylski:
User authentication with GitHub functionality is added according to the web application flow: https://docs.github.com/en/free-pro-team@latest/developers/apps/authorizing-oauth-apps.
Note: to receive a code to exchange it for access token, changes in app.tsx were made, however they should be moved in other place in the future as navigation is not yet done.

Added function to create new repository. The new repository is named "WEB_PUB + <project_name>.
Added function to get all public and private repositories. Function returns all repositories, which starts with "WEB_PUB".

## Linked issues
closes #20 #23

- [x] I have linked an issue (use e.g. `closes #20 #23 `)


src/main/git/gitAuthorization.ts
```diff
@@ -0,0 +1,32 @@
...
+export async function postAccessToken(codeParam : string) : Promise<AxiosResponse> {
+    return await axios.post(buildAccessTokenEndpoint(codeParam)).then(data => {
+        const search = data.data.substring(0);
+        const result = JSON.parse('{"' + search.replace(/&/g, '","').replace(/=/g,'":"') + '"}', function(key, value) { return key===""?value:decodeURIComponent(value) })
+        return result.access_token;
+    });
+}
+
+function buildAuthorizeEndpoint() : string {
+    let clientIdParam = "?"+ EndPointParameters.CLIENT_ID + "=" + process.env.GITHUB_CLIENT_ID;
```
lukasz-starosta, line 23: just a suggestion that JS has some built in functionalities to handle URL params. no need to change right now, but worth knowing:  
https://developer.mozilla.org/en-US/docs/Web/API/URL
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

src/renderer/components/app/app.tsx
```diff
@@ -4,17 +4,43 @@ import '../description/description';
 import Description from '../description/description';
 import { Provider } from 'react-redux';
 import { configStore } from '../../../shared/configureStore';
+import {postAccessToken} from "../../../main/git/gitAuthorization";
 
+require('dotenv').config();
+
+console.log(process.env);
```
lukasz-starosta, line 11: think we don't need the log here

src/renderer/components/app/app.tsx
```diff
@@ -4,17 +4,43 @@ import '../description/description';
...
+}
+
+class App extends React.Component <AppProps>{
+
+    async componentDidMount(): Promise<void> {
+        /** piece of code to get code after authorizing, code is a param of main page after making callback,
+        *   then having code it is exchanged for access token
+
+         */
+        let search = window.location.search;
```
lukasz-starosta, line 26: the URLSearchParams is especially useful here

emiliamarkowska (approved):

PR #29, opened 2020-11-26
Refactor user authentication

karczewskiPiotr:
## Description
User auth has been refactored to resolve security vulnerabilities and execute on the main process. Basic functionality of `UserLogin` wrapper is also in place so user has to log in before proceeding to the app.

There is a lack of automatic sign in for session persistency. That could probably be achieved by storing the `code` from GitHub API somewhere and requesting `access_token` on application mount or perhaps running authorization as it works based on cookies.

In order to have this run in your environment you have to create a `.env` file with:
```
GITHUB_CLIENT_ID=value
GITHUB_CLIENT_SECRET=value
```

Those can be acquired from creating a GitHub OAuth App or harassing @emiliamarkowska for her variables. 💁‍♀️  Probably good to mail Krzysztof about getting them from his organization @emiliamarkowska.
## Correct PR Checklist

### DIFF
--- a/src/main/git/gitTypes.ts
+++ b/src/main/git/gitTypes.ts
@@ -10,7 +10,7 @@ export interface File {
 }
 
 export enum Url {
-  REDIRECT_URI = 'http://localhost:3000/main_window',
+  REDIRECT_URI = 'http://localhost/main_window',
   AUTHORIZE_URL = 'https://github.com/login/oauth/authorize',
   ACCESS_TOKEN_URL = 'https://github.com/login/oauth/access_token',
 }
--- a/src/main/git/gitTypes.ts
+++ b/src/main/git/gitTypes.ts
@@ -1,23 +1,23 @@
 export interface Repository {
-    name: string,
-    author: string,
-    url: string
+  name: string;
+  author: string;
+  url: string;
 }
 
 export interface File {
-    filename: string,
-    path: string
+  filename: string;
+  path: string;
 }
 
 export enum Url {
-    REDIRECT_URI = "http://localhost:3000/main_window",
-    AUTHORIZE_URL = "https://github.com/login/oauth/authorize",
-    ACCESS_TOKEN_URL = "https://github.com/login/oauth/access_token"
+  REDIRECT_URI = 'http://localhost:3000/main_window',
+  AUTHORIZE_URL = 'https://github.com/login/oauth/authorize',
+  ACCESS_TOKEN_URL = 'https://github.com/login/oauth/access_token',
 }
 
 export enum EndPointParameters {
-    CLIENT_ID = "client_id",
-    REDIRECT_URI ="redirect_uri"
+  CLIENT_ID = 'client_id',
+  REDIRECT_URI = 'redirect_uri',
 }
 
-export const WEB_PUB_REPO_NAME = "WEB_PUB"
+export const WEB_PUB_REPO_NAME = 'WEB_PUB';
--- a/src/renderer/components/RouterComponents/Description/Description.tsx
+++ b/src/renderer/components/RouterComponents/Description/Description.tsx
@@ -1,16 +1,16 @@
 import React from 'react';
-import './description.scss';
+import './Description.scss';
 import { useDispatch, useSelector } from 'react-redux';
 import {
   example,
   exampleInMain,
   exampleLocal,
   selectCurrentUser,
-} from '../../../shared/slices/currentUserSlice';
+} from '../../../../shared/slices/currentUserSlice';
 import {
   installGatsbyCLI,
   selectGatsbyInstallStatus,
-} from '../../../shared/slices/gatsbyInstallSlice';
+} from '../../../../shared/slices/gatsbyInstallSlice';
 
 // TODO: delete exemplary buttons from here
 const Description = () => {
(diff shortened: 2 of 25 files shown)


karczewskiPiotr:
Okay, I implemented session persistency as well (a bit hacky perhaps but works nicely) in the last commit. So don't mind the initial message about it.

lukasz-starosta:
Please remember to document the fact that we need those .env variables, they are easy to forget

src/main/git/gitTypes.ts
```diff
@@ -1,23 +1,23 @@
...
-    path: string
+  filename: string;
+  path: string;
 }
 
 export enum Url {
-    REDIRECT_URI = "http://localhost:3000/main_window",
-    AUTHORIZE_URL = "https://github.com/login/oauth/authorize",
-    ACCESS_TOKEN_URL = "https://github.com/login/oauth/access_token"
+  REDIRECT_URI = 'http://localhost:3000/main_window',
```
lukasz-starosta, line 13: I think this will crash if we build the app, because it would not be running on localhost:3000 anymore, am I right?  
Anyway, maybe we could just do a relative path redirect to `/main_window` instead? It is safer and won't crash if port ever changes

src/renderer/components/RouterComponents/Description/Description.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import './Description.scss';
+import { useDispatch, useSelector } from 'react-redux';
+import { selectCurrentUser } from '../../../../shared/slices/currentUserSlice';
+import {
+  installGatsbyCLI,
+  selectGatsbyInstallStatus,
+} from '../../../../shared/slices/gatsbyInstallSlice';
+
+// TODO: delete exemplary buttons from here
```
lukasz-starosta, line 10: This comment is redundant now

lukasz-starosta (commented):
Nice job, just 2 small comments 😃

src/main/git/gitTypes.ts
karczewskiPiotr: I have not built this app ever before so honestly don't have any idea how it behaves in production. I have a feeling that since it's a desktop app Electron will work it's magic and the callback in production will still be localhost just not sure if the port will match. Because what sort of URL would it be if not a localhost.

217853 (approved):

src/main/git/gitTypes.ts
karczewskiPiotr: @lukasz-starosta using `http://localhost/main_window` works in development as well so I'm going to default to this for now. There is almost no info on the web about any sort of `root_url` for Electron apps or how to handle this and doing realtive `/main_window` does not work unfortunately as it would be relative to GitHub's url so this way at least we're getting rid of the port that could cause problems.

Based on my observations it will probably be `http://localhost/main_window` in production as well as the app runs locally and `main_window` is simply the app name and the point of entry is defined by webpack here:
```json
"plugins": [
  [
    "@electron-forge/plugin-webpack",
    {
      "mainConfig": "./webpack.main.config.js",
      "renderer": {
        "config": "./webpack.renderer.config.js",
        "entryPoints": [
          {
            "html": "./public/index.html",
            "js": "./src/renderer/renderer.tsx",
            "name": "main_window" <----
          }
        ]
      }
    }
  ]
]
```
and then used like this:
```ts
declare const MAIN_WINDOW_WEBPACK_ENTRY: any;
declare const MAIN_WINDOW_PRELOAD_WEBPACK_ENTRY: any;
.
.
.
const mainWindow = new BrowserWindow({
    height: 600,
    width: 1200,
    webPreferences: {
      preload: MAIN_WINDOW_PRELOAD_WEBPACK_ENTRY,
      enableRemoteModule: true,
      nodeIntegration: true,
    },
  });

  mainWindow.loadURL(MAIN_WINDOW_WEBPACK_ENTRY); <----
```

Definitely would be good to test the packaged up at some stage though as these are just assumptions.

ProudBloom (commented):
Don't know how about You @karczewskiPiotr, but both with Adam we get a compilation error when launching an application. It probably has to do with async functions and axios when fetching an API (I think it's the POST method, but not sure). I attached a screen-shot below, that displays an error. Also note that the components are not loading properly, yet the window still opens. 

image

karczewskiPiotr:
> Don't know how about You @karczewskiPiotr, but both with Adam we get a compilation error when launching an application. It probably has to do with async functions and axios when fetching an API. I attached a screen-shot below, that displays an error. Also note that the components are not loading properly, yet the window still opens. 
> 
> image
> 

Do you have the environmental variables set up?

ProudBloom:
> I don't have errors, otherwise this PR would not be here. 😅
> 
> @ProudBloom do you have the environmental variables set up? The window should open and the first thing the app does is check if user is logged in so it makes a request to GitHub API. So if that request fails for example because of bad env variables nothing else will render as the app needs to finish auth checks before deciding if it should render login screen or the app.

Ok, so I created a new (2nd) OAuthApp now and it seems to be working now. I have no idea why it haven't before 😅
Maybe I overlooked something or pasted a wrong URL.

ProudBloom (approved):

lukasz-starosta (approved):

PR #31, opened 2020-12-01
Add config manager

msuliborski:
## Description
Added createConfigFile function 
Added deleteConfigFile function
Added modify functions to each config attribute
Added getters functions to each config attribute

## Linked issues
Publication configuration file #3

## Correct PR Checklist
- [x] I have linked an issue (Publication confiration file #3`)


src/main/node/config-manager.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+}
+
+export function createConfigFile(path: string, name: string, description: string, collaborators: Colaborator[], 
+                          packageManager: string, tag: string): boolean {
+  if (isDirectory(path)) {
+    let configContent = { 
+      "name": name,
+      "description": description,
+      "collaborants": collaborators,
```
lukasz-starosta, line 20: There is no such word as `collaborant` I think :P

src/main/node/config-manager.ts
```diff
@@ -0,0 +1,152 @@
...
+        fs.unlinkSync(path + '/vn-pub.conf');
+        return true;
+      }
+    } catch(err) {
+      console.error(err);
+    }
+  return false;
+}
+
+export function modifyName(path: string, newName: string): boolean {
```
lukasz-starosta, line 44: Don't you think this function and other `modify` functions have a lot of code in common? If the filename changes then we have to remember to change it in every function. Same applies to other forms of handling the JSON

karczewskiPiotr (commented):
Why do we need those getter functions? Json should be read as a whole and pushed to Redux state eventually so there is no need to get individual fields, just the whole json read to a JavaScript object.

I think deleting the config file in each modification is unnecessary, could just write to it. Also, assuming the modifier functions write the whole new json to the file every time, we probably can only have one function taking in a JavaScript object of the config as a whole and write it to the file. Seems like it will provide the same functionality and be less troublesome to use

lukasz-starosta (approved):

karczewskiPiotr (approved):

PR #42, opened 2020-12-07
Log out functionality and auth improvements

### LINKED ISSUE #37: Integrate GitHub login/logout with our auth screen
## Description
User should be able to:
- log in by clicking the button
- receive error messages if something went wrong, the app must not crash
- log out after they have logged in

## Correct issue checklist:
- [x] I have specified the environment
- [x] I have added a Milestone (if applicable)

karczewskiPiotr:
## Description
This PR implements log out functionality, improves session handling and refactors auth.

In case of no environmental variables present GitHub will basically show this instead of log in form. 

image

There is no real good way of validating env variables as validating their presence does nothing for us, We'd need to check against GitHub if they are valid but there is no such functionality. So just make sure you have variables as the end user is bound to have them after deployment anyway.
## Linked issues
closes #37 

## Correct PR Checklist

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -3,6 +3,8 @@ import { configStore } from '../shared/configureStore';
 declare const MAIN_WINDOW_WEBPACK_ENTRY: any;
 declare const MAIN_WINDOW_PRELOAD_WEBPACK_ENTRY: any;
 
+require('dotenv').config();
+
 // Handle creating/removing shortcuts on Windows when installing/uninstalling.
 if (require('electron-squirrel-startup')) {
   app.quit();
--- a/src/renderer/components/NavigationBar/NavigationBar.tsx
+++ b/src/renderer/components/NavigationBar/NavigationBar.tsx
@@ -7,6 +7,7 @@ import {
 } from '../../../shared/slices/currentViewSlice';
 import './NavigationBar.scss';
 import { COMPONENTS } from '../../constants/Components';
+import { terminateSessionAsync } from '../../../shared/slices/currentUserSlice';
 
 const TOP_BUTTONS: COMPONENTS[] = [
   COMPONENTS.PROJECT,
@@ -67,7 +68,7 @@ const NavigationBar = () => {
         {renderListOfButtons(BOTTOM_BUTTONS)}
         <button
           className='navbar__button'
-          onClick={() => console.log('TODO: implement logout')}
+          onClick={() => dispatch(terminateSessionAsync())}
         >
           <div className='navbar__button__icon'>L</div>
           <span className='navbar__button__text'>Log out</span>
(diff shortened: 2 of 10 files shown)


217853 (approved):

src/main/index.ts
```diff
@@ -3,6 +3,8 @@ import { configStore } from '../shared/configureStore';
 declare const MAIN_WINDOW_WEBPACK_ENTRY: any;
 declare const MAIN_WINDOW_PRELOAD_WEBPACK_ENTRY: any;
 
+require('dotenv').config();
```
lukasz-starosta, line 6: Is this still needed?

src/renderer/components/NavigationBar/NavigationBar.tsx
```diff
@@ -67,7 +68,7 @@ const NavigationBar = () => {
         {renderListOfButtons(BOTTOM_BUTTONS)}
         <button
           className='navbar__button'
-          onClick={() => console.log('TODO: implement logout')}
+          onClick={() => dispatch(terminateSessionAsync())}
```
lukasz-starosta, line 71: Will this redirect us to the login screen too?

src/renderer/components/RouterComponents/Description/Description.tsx
```diff
@@ -26,19 +15,12 @@ const Description = () => {
...
-      <button
-        onClick={() =>
-          dispatch(displayUserData(currentUser.auth.accessToken.value))
-        }
-      >
-        Display user data
-      </button>
       <p>
-        Welcome <b>{currentUser.data.nick}</b>!
+        Welcome <b>{currentUser.data?.nick || ''}</b>!
```
lukasz-starosta, line 19: The correct syntax for optional chaining would be `currentUser.data?.nick ?? ''`

lukasz-starosta (commented):
Please respond to the issues in comments, I think it's alright just want to make it clear

src/main/index.ts
karczewskiPiotr: yep, renderer and main don't share them. previously env variables were passed down from renderer in functions calling API which just makes them have more params than necessary

src/renderer/components/NavigationBar/NavigationBar.tsx
karczewskiPiotr: yes 👌

src/renderer/components/RouterComponents/Description/Description.tsx
karczewskiPiotr: Cool stuff, didn't know about nullish coalescing operator before. 
Either way this was just placeholder when working and isn't really necessary as data is already there when this screen is rendered and in the end will be removed so won't bother.

lukasz-starosta (approved):

PR #43, opened 2020-12-07
Refactor view subview

### LINKED ISSUE #36: Refactor routing to have both the left and right screen separately
## Description
We should properly style and handle the two screens as seen below:
image
they should have some proper routing mechanism so that there is one gray background always in place and we only change the content.

## Correct issue checklist:
- [x] I have specified the environment
- [x] I have added a Milestone (if applicable)

lukasz-starosta:
## Description
There is now another component - they grey box on the right side of the screen:
image
It is hidden by default when we change the `View` and it can be shown by for example clicking a project on the projects list.
To show the `Subview` component, we have to dispatch an action with subview we want to show and props.

## Linked issues
closes #36 

## Correct PR Checklist
- [x] I have assigend a Milestone


src/renderer/components/RouterComponents/ProjectsList/IProject.ts
```diff
@@ -4,8 +4,9 @@ export interface IProject {
   id: number;
   image: string;
   name: string;
-  date_creation: Date;
-  date_edition: Date;
+  // Dates have to be serializable, cannot be of type Date
```
lukasz-starosta, line 7: @Eleara because of this change, the sorting in ProjectsList does not work properly. It may be enough to simply do `new Date(date_creation)` before comparing these dates, but I did not want to mess up your algorithm.

karczewskiPiotr (approved):

PR #45, opened 2020-12-08
Created directory selection view

### LINKED ISSUE #27: Implement project's directory choice view
## Description
Refer to this part:
image

## Correct issue checklist:

217868:
## Linked issues
closes #27 

## Correct PR Checklist

### DIFF
--- a/src/renderer/components/RouterComponents/Files/Directory/Directory.scss
+++ b/src/renderer/components/RouterComponents/Files/Directory/Directory.scss
@@ -1,56 +1,57 @@
 .directory {
-    height: 5.2rem;
-    display: flex;
-    align-items: center;
-    margin-top: 2rem;
+  height: 5.2rem;
+  display: flex;
+  align-items: center;
+  margin-top: 2rem;
 
-    width: 90%;
-    cursor: pointer;
+  width: 90%;
+  cursor: pointer;
 }
 
 .directory:hover {
-    >.directory__icon, .directory__content {
-        background-color: #8a97a1;
-    }
+  > .directory__icon,
+  .directory__content {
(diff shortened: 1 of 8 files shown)


lukasz-starosta:
Remember to format this with .prettierrc, you can run `yarn prettier --write --config .prettierrc ./src` when in project root directory, that should do the job

karczewskiPiotr:
> Remember to format this with .prettierrc, you can run `yarn prettier --write --config .prettierrc ./src` when in project root directory, that should do the job

There's a script in the project for simplicity so you can just run `yarn format`.

lukasz-starosta (approved):
I think it's fine, just format it so CI passes

lukasz-starosta (approved):

217853 (approved):

Eleara (approved):

karczewskiPiotr:
@217868 Not entirely sure but it seems like this may not be in the correct place of the app and in the wrong order of the app's lifecycle. I think it maybe doesn't do exactly what design team intended? From wireframes it looks like it's just a one time thing on the first log in but then again kind of loops. Maybe could convert this to WIP and consult the design team before merging?

217868:
> @217868 Not entirely sure but it seems like this may not be in the correct place of the app and in the wrong order of the app's lifecycle. I think it maybe doesn't do exactly what design team intended? From wireframes it looks like it's just a one time thing on the first log in but then again kind of loops. Maybe could convert this to WIP and consult the design team before merging?

I propose to merge it for now and modify this solution to being dependend on app settings after they are implemented. Then, we can display that screen only if these settings are absent in the app.

PR #93, opened 2021-10-29
Issue 79 - solved

### LINKED ISSUE #79: Define the structure for new project input data
Based on the screens from XD (_10 New project 1-6_) define an interface of the data structure that will hold the user input data in redux store. This needs to be done so that the back-end team can start working on functions depending on the input data. This way they will be able to mock it and don't need to wait for the actual UI to function properly.

JKusiak:
## Description
The publicationsSlice has been extended to contain reducers that operate on the data necessary for creating and managing publications. The Publication interface has been extended with new fields derived from user interface mock-ups.

The only field of the Publication interface that required a separate reducer was the collaborators array. Its inclusion into the common reducer that changes other fields would cause an issue with index access writes on union of keys, solving which would make the code less readable.

The solution has been tested by dispatching actions on a frontend button and ensuring proper state changes in the console.

## Linked issues
closes #79 

## Correct PR Checklist
- [x] I have linked an issue


src/shared/redux/slices/publicationsSlice.ts
```diff
@@ -34,24 +61,73 @@ const publicationsSlice = createSlice({
...
+    },
+    deleteCollaborator: (
       state: PublicationList,
-      action: PayloadAction<[PublicationListElement]>
+      action: PayloadAction<CollaboratorListModification>
     ) => {
-      state.list = action.payload;
+      const chosenPubIndex = state.list.findIndex(
+        (publication) => publication.id === action.payload.pubId
+      );
```
karczewskiPiotr, line 110: Can we standarize the `action.payload.id` / `action.payload.pubId` in the actions so the naming convention is the same?

src/shared/redux/slices/publicationsSlice.ts
```diff
@@ -1,31 +1,58 @@
...
+  id: string;
+  field: keyof PublicationPrimitives;
+  value: string;
+};
+
+// type created to define payload structure for actions on
+// publication collaborators
+type CollaboratorListModification = {
+  pubId: string;
+  value: Collaborator | string;
```
karczewskiPiotr, line 42: The typing for this action might be a little confusing. It will allow the developer to pass in a value of string for when he wants to add a collaborator (not only when deleting) and therefore just push a string to collaborators array. Perhaps make it into a generic.

```ts
interface CollaboratorListModification<T> {
  id: string;
  value: T
}
```

Then you can pass it in to the action definitions as `CollaboratorListModification<Collaborator>` and `CollaboratorListModification<string>`

src/shared/redux/slices/publicationsSlice.ts
```diff
@@ -1,31 +1,58 @@
-import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { createSlice, current, PayloadAction } from '@reduxjs/toolkit';
```
karczewskiPiotr, line 1: The `current` import here can be removed ([eslint](https://github.com/vnLab-Lodz/PubLab/runs/4065171538?check_suite_focus=true#step:4:41)).

karczewskiPiotr (changes requested):
ESlint has found some issues in the PR, which can be seen [here](https://github.com/vnLab-Lodz/PubLab/runs/4065171538?check_suite_focus=true#step:4:41) (since for some reason the checks don't show up correctly on this PR). You only need to take care of the ones with `error` status.

karczewskiPiotr:
FYI: I have also merged the `main` branch into this one for the updated `eslint` config.

src/shared/redux/slices/publicationsSlice.ts
```diff
@@ -1,57 +1,113 @@
...
-  description: string;
+type Publication = {
+  id: string;
   dirPath: string;
+  publicationName: string;
+  description: string;
+  collaborators: Collaborator[];
+  packageManager: string;
+  useSass: string;
+  useTypescript: string;
```
karczewskiPiotr, line 18: Just noticed now. 🤔 Wouldn't it make more sense for these two properties to be `boolean`?

src/shared/redux/slices/publicationsSlice.ts
```diff
@@ -1,57 +1,113 @@
...
+      action: PayloadAction<Publication[]>
+    ) => action.payload,
     addPublication: (
-      state: PublicationList,
-      action: PayloadAction<PublicationListElement>
+      state: Publication[],
+      action: PayloadAction<Publication>
+    ) => {
+      state.push(action.payload);
+      console.log(current(state));
```
karczewskiPiotr, line 52: I do not think we need these logs at all. The project comes in with [Redux dev tools](https://github.com/zalmoxisus/redux-devtools-extension) built into the development environment as set up in #86. The dev tools give a lot more control and information than the output of a log into console. 

#### Small preview

image

karczewskiPiotr (changes requested):
Great job overall! 🎇 Just last two minor things and this PR is going to be ready for merging.

PR #102, opened 2021-11-02
added imagepicker component

### LINKED ISSUE #69: Image picker
The `ImagePicker` component should take in the following props:

- `image` - the image src to be displayed or `undefined` 
- `onClick` - function to run when the image picker is pressed
- `error` - `boolean` that specifies whether the picker should render in error variant

With this approach the logic of picking an image can be externalised and component can be developed in storybook safely.

### Default

image

### Error

image

### With image selected

image

224107:
## Description
created imagepicker component using storybook
## Preview
Zrzut ekranu 2021-11-02 141425

&

fixes according to piotrs comments

## Linked issues

closes #69  

## Correct PR Checklist

- [x] I have linked an issue 


src/renderer/components/ImagePicker/ImagePicker.tsx
```diff
@@ -0,0 +1,26 @@
...
+      </div>
+    ) : (
+      <div className='img'>
+        <div className='imgplus' />
+      </div>
+    );
+  } else {
+    return <img className='img selected' src={image} />;
+  }
+};
```
karczewskiPiotr, line 24: I believe the component can be cleaned up a little by refactoring the if statement and using a function to generate the class name based on the `error` value. You can return the `img` if the `image` prop is present and then have default return with the classes generated from a function like so:

```tsx
const ImagePicker: React.FC<Props> = ({ image, error }) => {
  if (image)
    return <img className='img selected' src={image} alt='Example alt' />;

  const makeClass = (base: string): string => (error ? `${base} error` : base);

  return (
    <div className={makeClass('img')}>
      <div className={makeClass('imgplus')} />
    </div>
  );
};
```

The `onClick` prop also needs to be called when the element is pressed so it can be passed to an `onClick` event of the `div`.

```tsx
const ImagePicker: React.FC<Props> = ({ image, error, onClick }) => {
  // ... other code

  return (
    <div
      className={makeClass('img')}
      onClick={onClick}
    >
      <div className={makeClass('imgplus')} />
    </div>
  );
};
```

With the addition of an `onClick` handler comes a couple of things though. The element should define a role if it has event listeners so that needs to be included. It is a good practice to take care of accessibility and therefore have a keyboard event accompany the `onClick`. If that is done then also the element should be focusable. And if it is focusable it should also have title. Therefore, this leads to something along the lines of the below: 

```tsx
const ImagePicker: React.FC<Props> = ({ image, error, onClick }) => {
  // ... other code

  return (
    <div
      aria-label='image picker' // title for the accessible element
      tabIndex={0} // makes the element focusable with keyboard
      role='button' // see possible roles here -> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles
      className={makeClass('img')}
      onClick={onClick}
      onKeyPress={({ key }) => key === 'Enter' && onClick()} // allows the user to fire the onClick action with the press of Enter (when focused)
    >
      <div className={makeClass('imgplus')} />
    </div>
  );
};
```

karczewskiPiotr (changes requested):

karczewskiPiotr:
@224107 you can also inspect the output of ESlint [here](https://github.com/vnLab-Lodz/PubLab/runs/4085662011?check_suite_focus=true).

karczewskiPiotr (approved):

Issue #84, opened 2021-10-21
Implement app settings view


karczewskiPiotr:
The task of implementing this view can be split into couple key features.

- [x] #96
- [x] #97
- [x] #95
- [x] #98
- [x] #100
- [ ] The switches (left out for now)
- [x] #99
- [ ] Final step: implementing the view using proper components

### 0. The redux slice

Seems like the slice for the settings of the app has been previously implemented as `configurationSlice`. It might make more sense to refactor it into a `settingsSlice` so that the naming matches the type of data stored.

### 1. App version / Update

The current version should come from the `package.json` but can be stored in the `configurationSlice` (_see point above_) when the app starts. The flag that specifies whether the app is the latest version (eg. `isLatestVersion`) should also be put in the slice as it will allow for easier visual testing of the view.

The UI elements for notifying the user about new update should be implemented. The elements should not be visible if the app is the latest version.

> The functional side of updates will come in at a later stage of the development in the form of one of the strategies described [here](https://www.electronjs.org/docs/latest/tutorial/updates).

### 2. Internationalization

The language picker itself will be based on the #66 so should be available for use here.

The thing that needs implementing is the internationalization itself. The [react-i18next](https://react.i18next.com/) will probably be our best bet. We are aiming for Polish and English locales to be included in the project so those are the only two that need to be taken into consideration.

The language picker should then be the way to switch between the active locale.

### 3. The default directory for PaaW projects

The directory picker itself will be implemented in #70, therefore it should also be available for use.

The functional side of it should be implemented. User should be able to pick a directory here.

### 4. Notifications

Nothing more than the UI needs to be implemented for now. We will need to brainstorm what intervals for notifications should be available.

### 5. The switches

I think we can leave them out for now. They actually seem a little out of place as they are project based so we should move them to _Project Settings_.

### 6. Button / Data persistency

The button will be implemented in #65. 

When pressed, all the settings data stored locally in the view should be saved to the redux store and persisted as a json file on the local machine. This local state is important as we do not want the settings changes to take effect immediately but after the use presses the button.

The data persisted in json file should also be read on the app start and injected into the redux store.

###  7. Final step

Since components to be used in the view depend on another issues, the view itself will be implemented last. Previous subissues will be resolved only by implementing functionality and bare-minimum, placeholder UI.

## Preview

Refer to _11 App Settings_ in XD project for more details.

[...]

PR #104, opened 2021-11-05
Internationalization

### LINKED ISSUE #95: Internationalization
The language picker itself will be based on the #66 so should be available for use. A placeholder will be used until that is resolved.

The thing that needs implementing is the internationalization itself. The react-i18next will probably be our best bet. We are aiming for Polish and English locales to be included in the project so those are the only two that need to be taken into consideration.

The language picker should then be the way to switch between the active locale.

LostInStatic:
## Description

[`react-i18next`](https://react.i18next.com/) library was added and is ready to use throughout the UI code. 

So far the assumptions is that translations are kept in single files (one for every language). However, it is possible to split them (or even put them next to their respective components) if we decide it would improve workflow. 

Setting the `strict` flag in `tsconfig.json` was necessary for correct typing of the translation function (to provide a warning if a given key does not exist in translation file). All new warnings due to introduced rules were resolved.

The UI in `AppSettings.tsx` is a functional placeholder meant for demonstration and testing purposes - the proper implementation will be made later as described in #84

## Linked issues

Resolves #95


src/renderer/views/AppSettings/AppSettings.tsx
```diff
@@ -1,6 +1,36 @@
...
+
+  function submitChanges() {
+    dispatch(setLocale(language));
+  }
+
+  return (
+    <div>
+      {t('AppSettings.title')}
+      <Select
+        label={t('AppSettings.language')}
```
karczewskiPiotr, line 24: For accessibility reasons it might be worth giving this component an `aria-label` or `title` prop. I think the `label` prop in MUI is not for accessibility but rather for some sort of internal layout magic. Rest of the accessibility props are handled by MUI internally (focusing, roles and other aria tags).

src/renderer/views/AppSettings/AppSettings.tsx
```diff
@@ -1,6 +1,36 @@
...
+        value={language}
+        onChange={(e) => setLanguage(e.target.value as 'en' | 'pl')}
```
karczewskiPiotr, line 26: Just to future proof this a little (in case more languages come in the future) you could change the typing to `as typeof language` so that you don't have to come in and add them manually when new locales are added.

src/renderer/views/AppSettings/AppSettings.tsx
```diff
@@ -1,6 +1,36 @@
...
+      >
+        <MenuItem value='en'>{t('AppSettings.language.english')}</MenuItem>
+        <MenuItem value='pl'>{t('AppSettings.language.polish')}</MenuItem>
```
karczewskiPiotr, line 29: Similar future proofing could be done here by importing the `resources` from `i18next.ts` (because i18n does not provide a list of all supported languages unfortunately; it will bait you with `i18n.languages` but that's not all the available locales) and mapping over the keys of that object to generate menu items. With that, the key of translation could change to `AppSettings.language.pl` so that you can use the `resources` object key to retrieve the translation.

It could perhaps be even better to have a `supportedLocales` object exported from the `i18next.ts`. Something like: 
```ts
export const supportedLocales = Object.keys(resources) as SupportedLangCode[];
```

Then you could map over that instead of importing the whole `resources`.

karczewskiPiotr (changes requested):
The future proofing of locales isn't super necessary but a nice thing to have I think. Insignificant changes but improve the qol so we should probably go for it. 😄

karczewskiPiotr:
@LostInStatic one more thing is that it would be good to add the `@types/react-i18next` and `@types/i18next` to the project as dev dependencies. The packages did not seem to come with built in types.

src/renderer/components/App/App.tsx
```diff
@@ -8,9 +8,16 @@ import Auth from '../Auth/Auth';
...
+import { selectCurrentLocale } from '../../../shared/redux/slices/settingsSlice';
+import i18next from '../../internationalisation/i18next';
 
 const store = configStore('renderer');
 
+observeStore(store, selectCurrentLocale, (langCode) =>
+  i18next.changeLanguage(langCode)
+);
+
 const App = () => (
```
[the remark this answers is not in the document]
LostInStatic: Good point!

src/renderer/views/AppSettings/AppSettings.tsx
LostInStatic: Did not pay a lot of attention to that as the view is just a placeholder now, but I guess it is worth to fix that, if only to not introduce a bad example.

src/renderer/views/AppSettings/AppSettings.tsx
LostInStatic: done

src/renderer/views/AppSettings/AppSettings.tsx
LostInStatic: done

LostInStatic:
> 
> 
> @LostInStatic one more thing is that it would be good to add the `@types/react-i18next` and `@types/i18next` to the project as dev dependencies. The packages did not seem to come with built in types.

AFAIK, they have built-in types. They are typed correctly on my end. Also, trying to `yarn add` them results in following messages:
```
warning @types/i18next@13.0.0: This is a stub types definition. i18next provides its own type definitions, so you do not need this installed.
warning @types/react-i18next@8.1.0: This is a stub types definition. react-i18next provides its own type definitions, so you do not need this installed.
```

Could you double-check on your side,  and if the issue persists provide me with your environment details?

karczewskiPiotr:
> > @LostInStatic one more thing is that it would be good to add the `@types/react-i18next` and `@types/i18next` to the project as dev dependencies. The packages did not seem to come with built in types.
> 
> AFAIK, they have built-in types. They are typed correctly on my end. Also, trying to `yarn add` the types results in following messages:
> 
> ```
> warning @types/i18next@13.0.0: This is a stub types definition. i18next provides its own type definitions, so you do not need this installed.
> warning @types/react-i18next@8.1.0: This is a stub types definition. react-i18next provides its own type definitions, so you do not need this installed.
> ```
> 
> Could you double-check on your side, and if the issue persists provide me with your environment details?

@LostInStatic ahh yes this makes sense, I think I might have not run`yarn install` on my machine while looking at the code. 🤦

karczewskiPiotr (approved):

PR #106, opened 2021-11-08
Refactor/issue 76

### LINKED ISSUE #72: New project initialization flow

### LINKED ISSUE #76: Validate the correctness of the `config-manager` functions
The functions are available [here](https://github.com/vnLab-Lodz/PubLab/blob/main/src/main/node/config-manager.ts).

The functions should at least allow to set all the fields defined in #3.

- [ ] The config name should be changed to `publab.config.json`.

lprakapovich:
## Description

Service config-manager.ts was refactored according to the suggestions and tested with JEST.  

## Linked issues

closes #76 

## Correct PR Checklist

- [x] I have linked an issue 


src/main/node/config-manager.ts
```diff
@@ -1,144 +1,116 @@
...
+  validatePath(
+    !isDirectory(path),
+    'Error while creating a config file: directory is not found'
+  );
+  if (
+    !configuration.publicationName ||
+    !configuration.packageManager ||
+    !configuration.collaborators
+  ) {
+    throw new Error('Name, package manager and collaborators are required');
```
karczewskiPiotr, line 18: I think that only the `publicationName` should not be empty as it's user input but other fields will be handled with defaults so no need to check them,

src/main/node/config-manager.ts
```diff
@@ -1,144 +1,116 @@
...
-    deleteConfigFile(path);
-    fs.writeFileSync(`${path}/${configFileName}`, newConfigContentJSON);
-    return true;
-  }
-  return false;
+export function updateTag(path: string, tag: string) {
+  const configurationToUpdate: Configuration = getConfigFile(path);
+  updateConfigField(configurationToUpdate, path, () => {
+    configurationToUpdate.tag = tag;
+  });
```
karczewskiPiotr, line 73: All the update functions could be shortened to one overridable function to make everything more concise:

```ts
export function updateConfigField(
  path: string,
  field: keyof Omit<
    Configuration,
    'useTypescript' | 'useSass' | 'collaborators'
  >,
  value: string
): void;
export function updateConfigField(
  path: string,
  field: keyof Pick<Configuration, 'useTypescript' | 'useSass'>,
  value: boolean
): void;
export function updateConfigField(
  path: string,
  field: keyof Pick<Configuration, 'collaborators'>,
  value: Collaborator[]
): void;
export function updateConfigField(path: string, field: any, value: any) {
  if (field === 'publicationName') validateNonEmptyNorNull(value);

  const configuration = getConfigFile(path);
  const updatedConfiguration = { ...configuration, [field]: value };

  fs.writeFileSync(
    `${path}/${configFileName}`,
    JSON.stringify(updatedConfiguration, null, 2)
  );
}
```

With this we have one implementation with different overloads so that the typings of field and value always match and we significantly cut down on the amount of code.

It also allows to get rid of `saveUpdatedConfiguration`, `validateNonEmpty` and previous `updateConfigField` helper functions.

src/main/node/config-manager.test.ts
```diff
@@ -1,21 +1,266 @@
...
+  it('collaborators property cannot be empty or null', () => {
+    createConfigFile(nodeDirectoryPath, config);
+    expect(() => {
+      tryToUpdateWithInvalidParam(nodeDirectoryPath, null, updateCollaborators);
+    }).toThrowError();
+    expect(() => {
+      tryToUpdateWithInvalidParam(nodeDirectoryPath, [], updateCollaborators);
+    }).toThrowError();
+  });
+});
```
karczewskiPiotr, line 227: With the simplification of update functions to just one `updateconfigFunction` the tests also need refactoring. They can become more concise and readable with less code obstructing the meaning of the test. To start it off this would be the three main overloads tested:

```ts
describe('updateConfigField', () => {
  beforeEach(() => createConfigFile(nodeDirectoryPath, config));

  it('correctly updates the publicationName', () => {
    updateConfigField(nodeDirectoryPath, 'publicationName', 'newName');
    expect(getConfigFile(nodeDirectoryPath).publicationName).toBe('newName');
  });

  it('correctly updates useTypescript', () => {
    updateConfigField(nodeDirectoryPath, 'useTypescript', false);
    expect(getConfigFile(nodeDirectoryPath).useTypescript).toBe(false);
  });

  it('correctly updates collaborators', () => {
    updateConfigField(nodeDirectoryPath, 'collaborators', [
      { githubUsername: 'test', id: 'test', role: 'test' },
    ]);
    expect(getConfigFile(nodeDirectoryPath).collaborators).toEqual([
      { githubUsername: 'test', id: 'test', role: 'test' },
    ]);
  });
});
```

src/main/node/config-manager.ts
```diff
@@ -1,144 +1,116 @@
...
-    const configContentJSON = fs.readFileSync(`${path}/${configFileName}`);
-    const configContent = JSON.parse(configContentJSON);
-    configContent.description = newDescription;
-    const newConfigContentJSON = JSON.stringify(configContent);
-    deleteConfigFile(path);
-    fs.writeFileSync(`${path}/${configFileName}`, newConfigContentJSON);
-    return true;
-  }
-  return false;
+export function getConfigFile(path: string): Configuration {
```
karczewskiPiotr, line 76: Would be good to have a small test for this function too.

src/main/node/config-manager.ts
```diff
@@ -1,144 +1,116 @@
...
+  validatePath(
+    !isDirectory(path) || !isPublication(path),
+    'Config file does not exist. Check the provided path'
+  );
+  const configContentJSON = fs.readFileSync(`${path}/${configFileName}`);
+  return JSON.parse(configContentJSON) as Configuration;
```
karczewskiPiotr, line 82: In case this line gives you errors after changing import type (since the `readFileSync` return `Buffer` and `parse` expects a `string`) you can just do `JSON.parse(configContentJSON.toString())`.

karczewskiPiotr (changes requested):
Good base to work with! We can refactor this a little to make the code more concise, readable and reusable.

lprakapovich:
Thanks for meaningful comments and code examples. It was very helpful! 👍🏻

karczewskiPiotr (dismissed):

karczewskiPiotr (approved):

Issue #63, opened 2021-10-15
Input / Text Field


karczewskiPiotr:
The Input component needs to be available in three variants: **dark**, **light** and **error**.

If MUI is used in the project (#58) it could be just a simple style override of the [Text Field](https://mui.com/components/text-fields/) component.

For a more detailed info on padding, font sizes, colors etc. have a look at the XD design file available on Teams.

### Dark

![image](https://user-images.githubusercontent.com/34404855/137501132-84370352-79c2-4ec3-99d0-d4bd8a9c0fc2.png)

### Light

![image](https://user-images.githubusercontent.com/34404855/137501221-1b1d98d7-f5de-414f-a011-484c86084567.png)

### Error

![image](https://user-images.githubusercontent.com/34404855/137511496-b6fc478c-c95c-427d-acf9-c647d11ad732.png)

PR #107, opened 2021-11-09
Added OurTextArea for issue 64

### LINKED ISSUE #64: Text area
The `TextArea` component needs to be available in three variants: **dark**, **light** and **error**.

If MUI is used in the project (#58) this issue could become irrelevant (depending on implementation of #63 ). If the `Input` would be a simple styled MUI `TextField` then it would be enough to pass multiline props to it as specified in the [docs](https://mui.com/components/text-fields/#multiline).

### Dark

image

### Light

image

### Error

image

aleksander-lemiesz:
## Description

Added OurTextArea component

## Preview

obraz

## Linked issues

closes #64


kmaciejewska:
I will just add, before too much work is performed that this component can be merged with #63 as it will differ by just one prop - multiline. Also, as we discussed the palette colors have to be referenced directly here  (e.g theme.palette.lightRed.main) and mui.d.ts declarations then become reduntant. 
However, that seems to be causing issues with StoryBook:
image

karczewskiPiotr:
> I will just add, before too much work is performed, that this component can be merged with #63 as it will differ by just one prop - multiline. Also, as we discussed, the palette colours have to be referenced directly here (e.g theme.palette.lightRed.main) and mui.d.ts declarations then become redundant.
> 
> However, that seems to be causing issues with StoryBook:
> 
> image

@aleksander-lemiesz There are two ways of using the `theme` here. 

### 1. Params of `styled` function

You can use the `theme` from the params of styled function:

```tsx
const StyledTextArea = styled(TextField)(({ type, theme: { palette } }) => {
  const { lightGray, black, lightRed } = palette;
  const color = type === 'light' ? lightGray.main : black.main;

  return {
    // styles  
  }
});
```

This however implies that you then need to wrap the story in the `ThemeProvider`, which is not ideal since you have to do it for every component then.

```ts
const Template: ComponentStory<typeof OurTextArea> = (args) => (
  <ThemeProvider theme={theme}>
    <CssBaseline />
    <OurTextArea {...args} />
  </ThemeProvider>
);
```

So it can be generalised by wrapping all stories with the `ThemeProvider` with the use of [decorators](https://storybook.js.org/docs/react/writing-stories/decorators#context-for-mocking). `.storybook/preview.js` needs to be updated to the following to achieve that:

```js
import { theme } from '../src/renderer/theme';
import { ThemeProvider } from '@mui/material/styles';
import { CssBaseline } from '@mui/material';

export const parameters = {
  actions: { argTypesRegex: '^on[A-Z].*' },
  controls: {
    matchers: {
      color: /(background|color)$/i,
      date: /Date$/,
    },
  },
};

export const decorators = [
  (Story) => (
    <ThemeProvider theme={theme}>
      <CssBaseline />
      {Story()}
    </ThemeProvider>
  ),
];

```

It also seems like `storybook` is using `emotion` under the hood which conflicts with the new v5 `emotion` styling (see [here](https://github.com/mui-org/material-ui/issues/24282)). This is what is responsible for the error @kmaciejewska. It can be fixed by removing `emotion` aliases from storybook webpack configuration by updating the `.storybook/main.js` with the following:

```js
module.exports = {
  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
  addons: [
    '@storybook/addon-links',
    '@storybook/addon-essentials',
    '@storybook/preset-scss',
  ],
  webpackFinal(config) {
    delete config.resolve.alias['emotion-theming'];
    delete config.resolve.alias['@emotion/styled'];
    delete config.resolve.alias['@emotion/core'];
    return config;
  },
};

```

### 2. Directly from `theme` (less advised since it removes theme switching abilities)

```tsx
import { theme } from 'wherever it is';

const StyledTextArea = styled(TextField)(({ type }) => {
  const { lightGray, black, lightRed } = theme.palette;
  const color = type === 'light' ? lightGray.main : black.main;

  return {
    // styles  
  }
});
```

src/renderer/components/OurTextArea/OurTextArea.tsx
```diff
@@ -0,0 +1,67 @@
...
+      color: 'lightRed',
+    },
+  },
+  // customise helperText for errors
+  '& .MuiFormHelperText-root': {
+    color: 'lightRed',
+  },
+}));
+
+const OurTextArea: React.FC<Props> = (props) => (
```
karczewskiPiotr, line 42: The name `OurTextArea` gives off huge mother Russia vibes. 😅 You can just change it to `TextArea`.

src/renderer/components/OurTextArea/OurTextArea.tsx
```diff
@@ -0,0 +1,67 @@
...
+    '& fieldset': {
+      borderRadius: '0px',
+      borderColor: type === 'light' ? lightGray.main : black.main,
+      color: type === 'light' ? lightGray.main : black.main,
+    },
+    '&.Mui-error': {
+      '& fieldset': {
+        borderRadius: '0px',
+        borderColor: lightRed.main,
+      },
```
karczewskiPiotr, line 26: You're using `ligthGray` and `black` and `lightRead` but they do not come from anywhere. They are not defined at all.

karczewskiPiotr (changes requested):
@aleksander-lemiesz the code does not build. You can follow the directions from the comment to make it work.

karczewskiPiotr (approved):

PR #108, opened 2021-11-10
App version

### LINKED ISSUE #97: App version / Update
The current version should come from the package.json but can be stored in the configurationSlice (see point above) when the app starts. The flag that specifies whether the app is the latest version (eg. isLatestVersion) should also be put in the slice as it will allow for easier visual testing of the view.

The UI elements for notifying the user about new update should be implemented. The elements should not be visible if the app is the latest version.

> The functional side of updates will come in at a later stage of the development in the form of one of the strategies described here.

LostInStatic:
## Description

Version is extracted from `app.getVersion()` on startup and compared to the latest version (currently hardcoded, awaiting implementation) using `[semver](https://www.npmjs.com/package/semver)`. No periodic checks for latest version were implemented (it is so far unclear how frequent they should be/what should trigger them).  

UI is a placeholder to showcase the functionality. It is easiest to test it by changing the version in `package.json` (1.0.1 is the mocked latest).

## Linked issues

Resolves #97


src/shared/redux/slices/settingsSlice.ts
```diff
@@ -1,15 +1,24 @@
...
 import { SupportedLangCode } from '../../../renderer/internationalisation/i18next';
 import { RootState } from '../rootReducer';
 
 type Settings = {
   defaultDirPath: string;
   currentLocale: SupportedLangCode;
+  versionDetails: {
+    version: string;
+    isUpToDate: boolean;
+  };
```
karczewskiPiotr, line 12: Since you have the `VersionDetails` interface imported anyway and use it as payload type in the `setVersionDetails` you can also use it as a type here.

karczewskiPiotr (approved):
All good apart from that one possibility to reuse the type so I'll just preemptively accept this PR (hoping the review won't go stale).

> No periodic checks for latest version were implemented (it is so far unclear how frequent they should be/what should trigger them).

Regarding this, we can probably just settle down for fetching the latest version when the app is started. I don't think there is much need for http pooling. One thing we may consider later on is allowing the user to manually check for updates via some button in app settings.

src/shared/redux/slices/settingsSlice.ts
LostInStatic: Fixed, unfortunately the review went stale

LostInStatic:
@karczewskiPiotr I have decided to refactor the `AppSettings.tsx` before it gets more bloated - I have simply merged the branch with this one, but if you want the PR history to also reflect that (in addition to the commit history), I'll make a full PR if you let me know.

src/renderer/views/AppSettings/subcomponents/AppUpdate.tsx
```diff
@@ -0,0 +1,30 @@
...
+      {versionDetails.isUpToDate ? (
+        ''
+      ) : (
+        <>
+          <Typography>
+            {t('AppSettings.updates.newVersionAvailable')}
+          </Typography>
+          <Button>{t('AppSettings.updates.update')}</Button>
+        </>
+      )}
```
karczewskiPiotr, line 25: @LostInStatic this is a bit awkward. If you want to render something or nothing then you can use the `&&` operator like so:

```ts
{!versionDetails.isUpToDate && (
  <>
    <Typography>
      {t('AppSettings.updates.newVersionAvailable')}
    </Typography>
    <Button>{t('AppSettings.updates.update')}</Button>
  </>
)}
```

karczewskiPiotr (changes requested):

src/renderer/views/AppSettings/subcomponents/AppUpdate.tsx
LostInStatic: Must've left that one over from my first draft of the UI. Fixed.

karczewskiPiotr (approved):

PR #109, opened 2021-11-12
Default directory setting

### LINKED ISSUE #70: Directory picker
The `DirectoryPicker` component could be made from the combination of the `Input` (#63 ) and `Button` (#65) components. It should have an `onChange` prop for text edition of the directory path and a `onClick` prop for a function that rund when the button is pressed. It should also take in a `value` prop 

It should most likely have the following props:
- `placeholder` - for value to be shown if `Input` is empty
- `value` - for the value to be displayed as current
- `onChange` - function that would run on the `Input` field change
- `onClick` - function that would run on the `Button` press

This way the component is not responsible for the displayed content and the logic of picking the directory and can be developed in isolation.

### Preview

image

### LINKED ISSUE #98: The default directory for PaaW projects
The directory picker itself will be implemented in #70, therefore it should also be available for use.

The functional side of it should be implemented. User should be able to pick a directory here.

LostInStatic:
## Description

Logic for changing the setting was implemented with a placeholder UI. 

## Linked issues

Resolves #98

### DIFF
--- a/src/renderer/views/AppSettings/subcomponents/DefaultDirSelect.tsx
+++ b/src/renderer/views/AppSettings/subcomponents/DefaultDirSelect.tsx
@@ -1,0 +1,31 @@
+import { Button, InputLabel, TextField } from '@mui/material';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+const { dialog } = require('electron').remote;
+
+interface Props {
+  defaultDirPath: string;
+  onChange: (newDirPath: string) => void;
+}
+
+export default function DefaultDirSelect({ defaultDirPath, onChange }: Props) {
+  const { t } = useTranslation();
+  return (
+    <>
+      <InputLabel id='dir-select-label'>
+        {t('AppSettings.defaultDirectory.description')}
+      </InputLabel>
+      <TextField label='dir-select-label' value={defaultDirPath} />
+      <Button
+        onClick={() => {
+          dialog
+            .showOpenDialog({ properties: ['openDirectory'] })
+            .then(({ filePaths }: any) => onChange(filePaths[0]));
+        }}
+      >
+        {t('common.change')}
+      </Button>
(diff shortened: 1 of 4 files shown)


src/renderer/views/AppSettings/subcomponents/DefaultDirSelect.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export default function DefaultDirSelect({ defaultDirPath, onChange }: Props) {
+  const { t } = useTranslation();
+  return (
+    <>
+      <InputLabel id='dir-select-label'>
+        {t('AppSettings.defaultDirectory.description')}
+      </InputLabel>
+      <TextField label='dir-select-label' value={defaultDirPath} />
```
karczewskiPiotr, line 19: I know this is mocked directory picker anyway but maybe this can be a good tip. 

The label here just displays the given string as the MUI label:

image

image

If you want to say, which element labels the field you can use the `aria-labelledby` attribute (see more [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute))

karczewskiPiotr (approved):

src/renderer/views/AppSettings/subcomponents/DefaultDirSelect.tsx
LostInStatic: Huh, I think I meant to use `labelId` as with `Select`. A mistake anyway, since the property does not exist on `TextField`s. Just to be clear, do you want me to merge anyway and just make sure it will be correct on the final UI implementation?

karczewskiPiotr: @LostInStatic yeah you can merge no bother 👍

PR #111, opened 2021-11-12
Persist settings

### LINKED ISSUE #99: Button / Data persistency

LostInStatic:
## Description

Clicking the save button will now write the settings in form of a  `publab-settings.json` file in the [`userData`](https://www.electronjs.org/docs/latest/api/app#appgetpathname) folder. File will be read on startup, unless it is non-existent.

## Linked issues

Resolves #99


src/main/index.ts
```diff
@@ -43,7 +49,13 @@ app.on('ready', async () => {
     await installDevToolsExtensions();
   }
   await createWindow();
+
+  const settings = readSettingsFile();
+  if (settings) {
+    mainStore.dispatch(setAllSettings(settings));
+  }
   mainStore.dispatch(setVersionDetails(getVersionDetails()));
+  observeStore(mainStore, selectAllSettings, writeSettingsFile);
```
karczewskiPiotr, line 58: The only gripe I have with this is that over time we might build up the amount of observeStores we have. I tried to play around with the actions fired on main thread and there might be a way to actually use them like we'd want.

If we'd change the `persistSettings` functions to accept the path as a param:

```ts
const settingsFileName = 'publab-settings.json';

export function writeSettingsFile(path: string, settings: Settings) {
  const settingJSON = JSON.stringify(settings, null, 2);
  writeFileSync(`${path}/${settingsFileName}`, settingJSON);
}

export function readSettingsFile(path: string): Settings | null {
  const fullPath = `${path}/${settingsFileName}`;
  if (existsSync(fullPath)) {
    const settingsJSON = readFileSync(`${path}/${settingsFileName}`).toString();
    return JSON.parse(settingsJSON);
  }
  return null;
}
```

and then created the action with `createAsyncActionMain` helper within the `settingsSlice`:

```ts
export const saveSettingsThunk = createAsyncActionMain<{
  path: string;
  settings: Settings;
}>('saveSettings', ({ path, settings }) => async (dispatch) => {
  dispatch(setAllSettings(settings));

  writeSettingsFile(path, settings);
});
```

we would then be able to fire it from the renderer code by getting the `app` from remote module:

```tsx
const path = require('electron').remote.app.getPath('userData');

const AppSettings = () => {
  const { t } = useTranslation();
  const [settings, changeSetting] = useReducer(
    handleSettingChange,
    useSelector(selectAllSettings)
  );

  function submitChanges() {
    dispatch(saveSettingsThunk({ path, settings }));
  }

  return (
    <div>
      // ... other code
      <Button onClick={() => submitChanges()}>{t('common.save')}</Button>
    </div>
  );
};
```

I feel like this is more inline with the idea behind using [electron-redux](https://github.com/klarna/electron-redux) package. It also means that the code, which performs the save of the settings is fired by the button press rather than by listening to changes of the store value.

What do you think @LostInStatic?

> We could also potentially create a set of `renderer/constants.ts` for things like this user path that could only be used in front-end code since they would utilise the remote package to get the app.

LostInStatic: The biggest issue I had with your proposed solution was the shift in responsibility for "knowing" the path to the UI component. However, this gave me a hunch about another solution to the problem. A module like this would allow us to safely (as both have the same type) reference `app`  in the code shared between processes (in this case, settings slice):
``` javascript
import { app as mainApp, remote } from 'electron';

const isRenderer = process && process.type === 'renderer';

const app = isRenderer ? remote.app : mainApp;
export default app;
```
Did a quick test, it seems to work as intended. 

I am still not fully convinced against the observer in this specific case. On one hand, I also don't like the idea of constantly monitoring the state for changes (especially taking into account the potential issues with deep comparison). On the other, handling in in the reducer means having to remember about saving the settings in every reducer or using only `setAllSettings` to handle every change. 

Taking all into consideration, I'm the least doubtful (even if by a small margin) about the solution involving calling `writeSettingsFile`  in every reducer - hopefully we will not need too many ; )

As that is he approach you also leaned towards I'll implement that, including the module described above.

karczewskiPiotr: @LostInStatic let'd do as you say for now. 👍 

I think we should have a brainstorm at the end of the year to evaluate the architecture of the app and whether `elecron-redux`'s aliased actions that can make code run on the main thread are actually helping us or weirdly bluring the border between responsibilities of renderer and main threads.

LostInStatic: Added two thunks to `settingsSlice` (file input and output), made changes (now takes partial settings object and spreads it with old state) to `setAllSettings` to cover most practical needs and abstracted file operations (so it can be reused) - now most information directly related to the app settings is contained in the slice.

I went ahead and converted the existing code to use new thunks, removed unnecessary exports and reducers. I also moved version details out of the slice (as there is a convenient way to access the information directly).

src/shared/redux/slices/settingsSlice.ts
```diff
@@ -60,7 +44,20 @@ export const selectDefaultDirPath = (state: RootState) =>
...
   state.appSettings.currentLocale;
 
-export const selectVersionDetails = (state: RootState) =>
-  state.appSettings.versionDetails;
+export const saveSettingsAsync = createAsyncActionMain<Partial<Settings>>(
+  'saveSettings',
+  (settings) => async (dispatch) => {
+    dispatch(setSettings(settings));
+    writeJSON(SETTINGS_FILE_PATH, settings);
+  }
```
karczewskiPiotr, line 52: @LostInStatic this will save the partial to the json file. In case of settings view it works all goof but in case of the lang changes on log in screen it will save only the `currentLocale` to the json.

persist

Perhaps using the `getState` can be of help here:

```ts
export const saveSettingsAsync = createAsyncActionMain<Partial<Settings>>(
  'saveSettings',
  (settings) => async (dispatch, getState) => {
    const nextSettings = { ...getState().appSettings, ...settings };
    dispatch(setSettings(nextSettings));
    writeJSON(SETTINGS_FILE_PATH, nextSettings);
  }
);
```

karczewskiPiotr (changes requested):

src/shared/redux/slices/settingsSlice.ts
LostInStatic: Right... 🤦‍♂️ Fixed.

LostInStatic:
I've stumbled upon another bug - state won't be re-read from the file on window reload. There is no reason for the user to do that now, but it might possibly introduce unexpected behavior in further development. I'll fix it.

karczewskiPiotr (approved):

PR #112, opened 2021-11-13
[Issue77] Config files modification

### LINKED ISSUE #77: Implement modification of PaaW project config files
The `package.json` and `gatsby-config.js` or `gatsby-config.ts` files need to be modified in order to include the details input by the user during the new project flow (#72).

BaranowskiBrt:
## Description
Functions allowing for changing the configs in the new project flow.
Blocked by #106 (I would have to make the same changes)

## Linked issues

Closes #77 

## Correct PR Checklist

- [x] I have assigned a Milestone (if applicable)


BaranowskiBrt:
@karczewskiPiotr I haven't implemented changing of `gatsby-config.js`, as in the templates it exports the typescript file, should the code from typescript config be changed and somehow written to the javascript file?

src/main/node/config-manager.ts
```diff
@@ -80,3 +83,66 @@ function validateNonEmptyNorNull(property: string): void {
...
+  if (!config.useSass) {
+    delete pckgJson.dependencies.sass;
+  }
+  if (!config.useTypescript) {
+    delete pckgJson.devDependencies.typescript;
+  }
+  fs.writeFileSync(packagePath, JSON.stringify(pckgJson));
+}
+
+export function modifyGatsbyConfig(path: string, config: Configuration) {
```
karczewskiPiotr, line 122: This one looks to work which is really cool. 🎊

karczewskiPiotr (changes requested):
> @karczewskiPiotr I haven't implemented changing of gatsby-config.js, as in the templates it exports the typescript file, should the code from typescript config be changed and somehow written to the javascript file?

All good, the current approach works great. Only the TypeScript version of the config needs changing since the js one simply exports it after transpilation. So no further changes regarding that need to happen.

Minor changes to te `modifyNodePackage` are needed to make it work but apart from that great job. 🥇

src/main/node/config-manager.ts
```diff
@@ -80,3 +83,63 @@ function validateNonEmptyNorNull(property: string): void {
...
+
+export function modifyNodePackage(path: string, config: Configuration) {
+  const packagePath = `${path}/${NODE_PACKAGE_PATH}`;
+  validatePath(
+    !isDirectory(path) || !fs.lstatSync(packagePath).isFile(),
+    "Couldn't find the package.json file."
+  );
+
+  // eslint-disable-next-line import/no-dynamic-require
+  const pckgJson = require(packagePath);
```
karczewskiPiotr, line 105: @BaranowskiBrt Seems like I did not add the comment to the last review before publishing it. But this should be

```ts
const pckgJson = JSON.parse(fs.readfileSync(packagePath).toString())
```
Otherwise the dynamic require will fail for files that are outside of the repository I think since they are not packaged before runtime. It's also good to keep this in line with how other files are accessed.

Sorry for the inconvenience. 🤦

karczewskiPiotr (changes requested):

karczewskiPiotr (approved):

PR #115, opened 2021-11-13
Loginscreen front2

### LINKED ISSUE #81: Implement the log in view
The view in question is called _01 log in/out 1_ in XD design. 

The `LoginComponent` should be modified to look like the design. The `authorizeGitHubUserAsync` action can be dispatched to log in the user just like currently done in the component.

For a more detailed info on padding, font sizes, colors etc. consult the XD design file available on Teams.

image

Patrycjaaaa:
## Description
Logging screen was finished by adding language buttons

## Linked issues

 closes #81

## Correct PR Checklist


src/renderer/components/LoginComponent/LoginComponent.tsx
```diff
@@ -1,32 +1,55 @@
...
+            }}
+          >
+            <FaGithub className='icon' /> <p>Log in with GitHub </p>
+          </Button>
+        </div>
+      </div>
+      <div className='footer'>
+        <Typography variant='h4'>
+          <p>
+            {' '}
```
karczewskiPiotr, line 44: You can get rid of all the `{' '}`.

src/renderer/components/LoginComponent/LoginComponent.tsx
```diff
@@ -1,32 +1,55 @@
...
+    <div className='background'>
+      <div className='language-box'>
+        <Button variant='text' className='language'>
+          {' '}
+          <p>English</p>{' '}
+        </Button>
+        <Button variant='text' className='language'>
+          {' '}
+          <p>Polski</p>{' '}
+        </Button>
```
karczewskiPiotr, line 22: You can use the `supportedLocales` object to render the buttons with code along the lines of this:

```tsx
const LoginComponent = () => {
  const { t } = useTranslation();

  return (
    <div className='background'>
      <div className='language-box'>
        {supportedLocales.map((locale) => (
          <Button key={locale} variant='text' className='language'>
            <p>{t(`AppSettings.language.${locale}`)}</p>
          </Button>
        ))}
      </div>
      // ... other code
    </div>
  );
};
```

This way it will render all possible locales that the app has as it could change in the future. This will also already intorduce a bit of internationalisation to the view.

karczewskiPiotr (changes requested):
Positioning things absolutely is in most cases not the right approach and same goes for this view. The very specific percentage values do not work across the board.

image

You can see how the elements do not end up in the centre when the view shrinks down. You could do `50%` values and `transform: translate(-50%, -50%)` for the absolutely positioned elements but it would be much better to just use flexbox in the .background element and just justify the elements with space between.

```css
display: flex;
justify-content: space-between;
flex-direction: column;
```

That way it is centered on all screen sizes.

karczewskiPiotr:
@Patrycjaaaa let me know when this is ready for rereview, thanks. 🙏

Patrycjaaaa:
@karczewskiPiotr could you review the code? It is improved, however ESlint doesn’t go through it and I don’t know how to fix it.

karczewskiPiotr (approved):
@Patrycjaaaa workloads are not run when there are merge conflicts so eslint didn't fire up with new commits.

PR #118, opened 2021-11-16
Node.js installation

kmieciakm:
## Description

Provides nodejs installation on Windows, Linux desktop, and Mac OS.

## Preview

not applicable

## Linked issues

kind of closes [73](https://github.com/vnLab-Lodz/PubLab/issues/73)

## Correct PR Checklist


src/main/node/node.ts
```diff
@@ -0,0 +1,101 @@
...
+  public static Current(): Platform {
+    return new Platform(type());
+  }
+
+  public IsSupported(): boolean {
+    const index = Platform.supportedPlatforms.findIndex(
+      (platform) => platform.os === this.os
+    );
+    return index > -1;
+  }
```
karczewskiPiotr, line 33: This can be done with the `Array.prototype.some` method:

```ts
public IsSupported(): boolean {
  return Platform.supportedPlatforms.some(({ os }) => os === this.os);
}
```

src/main/node/node.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  public NodeManagerInstallationCommand(): string {
+    if (this === Platform.WINDOWS) return 'winget install nvs';
+    if (this === Platform.LINUX || this === Platform.MAC)
+      return 'npm install -g n';
+    return '';
+  }
```
karczewskiPiotr, line 40: It would make sense semantically if this returned `string` or `undefined` when the platform does not match any supported ones.

src/shared/redux/slices/nodeInstallSlice.ts
```diff
@@ -26,16 +26,26 @@ const nodeCheckSlice = createSlice({
...
+      installNode()
+        .then(() => {
+          dispatch(nodeInstalled());
+        })
+        .catch(() => {
+          dispatch(nodeNotInstalled());
+        });
     }
   }
 );
```
karczewskiPiotr, line 51: You can utilize theh `try..catch` block here. 

```ts
export const installNodeJs = createAsyncActionMain<void>(
  'installNode',
  () => async (dispatch) => {
    dispatch(checkingInstallation());

    try {
      const installed = await checkForNode();

      if (!installed) await installNode();

      dispatch(nodeInstalled());
    } catch (error: any) {
      dispatch(nodeNotInstalled());
    }
  }
);
```

src/main/node/node.ts
```diff
@@ -0,0 +1,101 @@
...
+        resolve();
+      });
+      exec(platform.NodeInstallationCommand(), (error) => {
+        if (error) {
+          appendLog(`Node installation failed.`);
+          reject(error);
+        }
+        appendLog(`Finished installation of node.`);
+        resolve();
+      });
```
karczewskiPiotr, line 94: These two execs will run almost at the same time aren't they? The second one should be run if the first one finishes without an error.

karczewskiPiotr (changes requested):
> Not bad. The question is how will it age? So far I didn't put that in a code.

Given the extent of the [dist](https://nodejs.org/dist/) it seems like it will age well. If anything changes regarding this maintenance can always be undertaken. I'd say go for it.

package.json
```diff
@@ -131,6 +131,7 @@
     "electron-redux": "^1.5.4",
     "electron-squirrel-startup": "^1.0.0",
     "i18next": "^21.3.3",
+    "os": "^0.1.2",
```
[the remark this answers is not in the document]
kmieciakm: You are right! I have removed the dependency.

src/main/node/node.ts
kmieciakm: Realized as suggested.

src/main/node/node.ts
kmieciakm: Done.

src/shared/redux/slices/nodeInstallSlice.ts
kmieciakm: Done.

src/main/node/node.ts
kmieciakm: Sure, now they execute independently.
`platform.InstallNodeManager().then(() => platform.InstallNodeJs())`

src/main/node/node.ts
```diff
@@ -0,0 +1,124 @@
...
+      }
+      const message = `Node installation command unspecified.`;
+      reject(message);
+    });
+  }
+
+  private NodeManagerInstallationCommand(): string | undefined {
+    if (this === Platform.WINDOWS) return 'winget install nvs';
+    if (this === Platform.LINUX || this === Platform.MAC)
+      return 'npm install -g n';
```
karczewskiPiotr, line 71: @kmieciakm so we're not moving over to installation form dist in the end?

kmieciakm: Dodane

PR #120, opened 2021-11-18
Select

### LINKED ISSUE #66: Select
The `Select` component needs to be available in two variants: **dark** and **light**.

If MUI is used (#58) then the implementation of the select becomes easy as it's just styling. In other case there is a dependency of `react-select` library in the project from last year, which could be styled an used instead.

### Dark

image

image

 ### Light

Looks like it probably only differs by the outline color but double check in the XD design.

image

Happiness07:
## Description

Created Select according to AdobeXD

## Linked issues

closes #66 

## Correct PR Checklist


karczewskiPiotr (approved):
@Happiness07 overall a good approach! 🎊 

Selects are always a bit of a struggle to style in MUI so I made a few changes to make it fully work.

Issue #65, opened 2021-10-15
Button


karczewskiPiotr:
The button component has quite a few variants in the project. The button can differ in:
1. width
2. height
2. font weight
3. font size
4. text case
5. background color
6. outline color
7. text color

Because of all these factors there are two possible approaches to the implementation:
1. Have one `Button` component that takes in props, which allow to specify the look of the button. (a somewhat modular button)
2. Have multiple different button components exported. This would be fairly convenient if MUI is used in the project (#58) as it would be possible to just export styled versions of their `Button`.

For a more detailed info on padding, font sizes, colors etc. have a look at the XD design file available on Teams.

### Variants

- [x] Button with icon 
![image](https://user-images.githubusercontent.com/34404855/137507393-84e9fd71-0fb5-46cd-a31c-4fd8e6940ad1.png)

- [x] Button with text in sentence case text
![image](https://user-images.githubusercontent.com/34404855/137507439-b00d9d53-d232-4526-8ae3-9a1af9fc3620.png)

- [x] Button with light upper case text
![image](https://user-images.githubusercontent.com/34404855/137507498-960d8d7e-fdd0-4ce5-9052-608c6b7e969f.png)

- [x] Button with bold upper case text 
![image](https://user-images.githubusercontent.com/34404855/137507624-c1c768b1-b33e-4db4-8df2-2e4058959f7b.png)

- [ ] Button with an outline (**dark** and **light** variants) 
![image](https://user-images.githubusercontent.com/34404855/137508105-d53233c0-61ba-4898-8a89-9478094d3b90.png) 
![image](https://user-images.githubusercontent.com/34404855/137507951-c49b2de7-d8da-4b71-a40e-6c9836bc9477.png)

karczewskiPiotr:
In case a more singular approach is used for implementing the button/buttons it is possible to quickly create new issues for the checklist elements by pressing the small icon in the upper right corner so that the work on the component can be more granular.

![issues](https://user-images.githubusercontent.com/34404855/137509960-ba37d90a-fd26-4706-8cb7-ab0a5ec83b5d.gif)

PR #121, opened 2021-11-18
Issue70

### LINKED ISSUE #70: Directory picker

DziobakSz:
## Description

Couldn't use #65 component because it's not finished. 

## Linked issues

closes #70

## Correct PR Checklist


kmaciejewska (changes requested):
The function that fires on the Input change should probably also be passed as a prop since the parent component will most likely use the value - and right now there is no such possibility.

src/renderer/components/DirectoryPicker/DirectoryPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+  value?: string;
+  placeholder?: string;
+  onClick?: () => void;
+}
+
+const DirectoryPicker: React.FC<Props> = ({ value, placeholder, onClick }) => {
+  const [currentValue, setCurrentValue] = React.useState(value);
+
+  const handleDirectoryChange = (event: React.ChangeEvent<HTMLInputElement>) =>
+    setCurrentValue(event.target.value);
```
kmaciejewska, line 15: handleDirectoryChange in this case will just come as a prop, the same way you handled onClick. Thus, this piece of code becomes redundant

kmaciejewska (changes requested):

kmaciejewska (dismissed):

karczewskiPiotr (approved):

PR #122, opened 2021-11-18
Issue 74

### LINKED ISSUE #74: Implement function to pick the `PaaW` template based on technologies
Based on a `Publication`  (see type def in publicationSlice) configuration the template should be chosen. This basically means we need a function that will return a url to the template based on the use of `sass` and `typescript`. For now the logic could be:

* if typescript and sass enabled then return https://github.com/vnLab-Lodz/gatsby-starter-paaw-i18n
* otherwise return https://github.com/vnLab-Lodz/gatsby-starter-paaw-basic

We'll expand templates later on but no need for that yet.

matstolar:
## Description
Added function to get template URL based on technologies selected by the user.

## Linked issues

Closes #74 

## Correct PR Checklist


matstolar:
I've put this function in gatsbyGenerateProject reducer, as it will be probably only used when creating a new project. I don't think there's a need to create another component for this. I couldn't specify the publication to get value from, because the creation of publications is not yet implemented (I suppose).

karczewskiPiotr (approved):

Issue #129, opened 2021-11-25
Create view for adding collaborators


karczewskiPiotr:
For now the collaborators can be added to the list without a real verification if such an account on GitHub exists. The available roles should be a `programmer` and an `editor` (capitalize values for display).

Have a look at #125 to see where the elements should be put in order to render in the correct spot.

### Preview

![image](https://user-images.githubusercontent.com/34404855/143437705-c706f454-fa1d-4a24-a15c-d8f1da0b202e.png)

Issue #128, opened 2021-11-25
Create view for picking the project technologies


karczewskiPiotr:
By default both options can be toggled.

Have a look at #125 to see where the elements should be put in order to render in the correct spot.

### Preview

![image](https://user-images.githubusercontent.com/34404855/143437529-e0eb29e2-d68a-4e64-a162-cec8fb6ce3c4.png)

Issue #127, opened 2021-11-25
Create view for choosing package manager


karczewskiPiotr:
For now npm should be selected by default and yarn should be disabled.

Have a look at #125 to see where the elements should be put in order to render in the correct spot.

### Preview

![image](https://user-images.githubusercontent.com/34404855/143437379-3aef1b90-1188-4637-ba6b-61b1bac930ac.png)

Issue #126, opened 2021-11-25
Create view with form for general project info


karczewskiPiotr:
The form elements should be in error state if the values are not provided after the initial focus. Have a look at #125 to see where the elements should be put in order to render in the correct spot.

The next button should be disabled if the form has errors.

### Preview

![image](https://user-images.githubusercontent.com/34404855/143437098-25ecfeda-07d1-449d-a663-1c076f8187cc.png)

PR #144, opened 2021-11-29
Add and test addPublicationSlice, change slices names accordingly to …

### LINKED ISSUE #125: Create a way to handle the multi step form for project creation
Since the project creation views mentioned in #123 are essentially a multi step form, we need a way to handle all the value changes and keep track of the data as well as which part of the form should be rendered. 

The main view should therefore have the header and steps indicator and the components in the green zone should be rendered based on the current step.

The contents of the green zone will be implemented in other issues. This issue should only provide a way to render them.

image

JKusiak:
Creation of a new Redux slice that would handle adding a new project to the system.

## Description
The slice resembles the one responsible for loading data from the .config file. In order to distinguish between the two, as they serve different purposes, the naming convention has been adjusted to reflect the difference. The slice holds an initial empty state that gets updated as the data is introduced into the form.

## Linked issues
closes #125

## Correct PR Checklist


src/shared/redux/slices/addPublicationSlice.ts
```diff
@@ -0,0 +1,87 @@
...
+      value: string;
+    }
+  | {
+      field: keyof Pick<NewPublication, 'useTypescript' | 'useSass'>;
+      value: boolean;
+    }
+  | {
+      field: keyof Pick<NewPublication, 'step'>;
+      value: number;
+    };
```
basia-borkowska, line 35: How about structuring this type that way? 😄 I feel it's a bit more readable and takes up less space 😆 
```suggestion
type PublicationModification = {
  value: string | boolean | number;
  field: keyof (
    | Pick<NewPublication, 'useTypescript' | 'useSass'>
    | Pick<NewPublication, 'step'>
    | Omit<
        NewPublication,
        'useTypescript' | 'useSass' | 'collaborators' | 'step'
      >
  );
};
```

basia-borkowska (approved):
Looks good, great job 🥇 
I have one question to @karczewskiPiotr. Why doesn't the project have a separate file for `types`? 🤔  Since we use them throughout multiple different files (eg. `Collabolator` or `Publication`), wouldn't it be cleaner to extract them to one file and keep them there?

src/shared/redux/slices/addPublicationSlice.ts
karczewskiPiotr: @basia-borkowska it takes up less space but it does not have the same semantic meaning and therefore, does not unfortunately allow to achieve the correct typings. This typings are for matching the `value` type with a specific `field` type. In your suggestion the user can set a value of `boolean` to any field for example, which is not desired. Whereas in the original typings if the `value` is `boolean` the field can only be set to `useTypescript` or `useSass`.

@JKusiak has more insight on the topic if you'd like to inquire further.

karczewskiPiotr:
> Looks good, great job 🥇 I have one question to @karczewskiPiotr. Why doesn't the project have a separate file for `types`? 🤔 Since we use them throughout multiple different files (eg. `Collabolator` or `Publication`), wouldn't it be cleaner to extract them to one file and keep them there?

@basia-borkowska It would definitely be a great thing to have. It may sometimes be hard to decide what should be in such a global type scope so I didn't push for this for now. This is however planned for the end of December when me and @LostInStatic will sit down to evaluate the project architecturally, structurally and in terms of development environment.

karczewskiPiotr (changes requested):
Given that this PR is marked as closing #125 it seems to be lacking the UI for the multi step form with the outlet for specific steps that will be implemented in #126, #127, #128 and #129.

@JKusiak will this be implemented in another PR, in which case the issue should be split, or simply missing from this PR?

karczewskiPiotr:
@JKusiak some merge conflicts seem to have creeped. In case of any issues you can ping @LostInStatic for some help since he's the author of the changes.

src/renderer/views/AddProject/AddProject.tsx
```diff
@@ -0,0 +1,114 @@
...
+  height: '60px',
+  border: 'none',
+  outline: 'none',
+  background: theme.palette.green.main,
+  fontSize: '15px',
+  fontWeight: 'bold',
+  '&:hover': {
+    cursor: 'pointer',
+  },
+}));
```
karczewskiPiotr, line 58: Hmm, why aren't we using the project's `Button` component here but instead styling new ones?

Also, In case of _css-in-js_ styling it is always good to extract them to another file like `style.ts` as well.

src/renderer/views/AddProject/AddProject.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <StyledButtonGreen
+            onClick={() => {
+              if (currentStep < 5) dispatch(increaseStep());
+            }}
+          >
+            NEXT
+          </StyledButtonGreen>
+        </div>
+      </StyledContainer>
+    </StyledBackground>
```
karczewskiPiotr, line 110: It would be great if the view behaved the same as the `AppSettings` one in terms of it's width and overall styling. You can have a look at it and use the same approaches there as the `ViewContent` component is reusable.

The changes could probably end up being as simple as just exchanging the `Styledbackground` for a `ThemeProvider` with an alt theme and the `StyledContainer` for the `ViewContent`.

karczewskiPiotr (changes requested):
Using the `altTheme`  is definitely the way to go for this as it will then match the design in XD. 

Internationalisation for the text would also be super handy for this component already.

karczewskiPiotr (approved):

PR #145, opened 2021-11-30
Theme refactor

LostInStatic:
## Description

A better implementation of themes. Most MUI components should have colors close to intended by default. Additionally, it is no longer necessary to specify whether a component is 'light' or 'dark' - it is controlled by `ThemeProvider` instead, so the component will match it's surroundings unless the color is overridden.

Main theme is 'light', as more components use it according to the design. By wrapping components in another `ThemeProvider` with `altTheme` passed to it you can change their color scheme. 

Storybook config was also updated to allow changing themes while previewing components.

### DIFF
--- a/src/renderer/components/ViewContent/ViewContent.tsx
+++ b/src/renderer/components/ViewContent/ViewContent.tsx
@@ -2,9 +2,9 @@ import { BoxProps } from '@mui/material';
 import React from 'react';
 import { BackgroundWrapper, ContentBox } from './style';
 
-const ViewContent: React.FC<BoxProps> = (props) => (
-  <BackgroundWrapper {...props}>
-    <ContentBox>{props.children}</ContentBox>
+const ViewContent: React.FC<BoxProps> = ({ children, ...rest }) => (
+  <BackgroundWrapper {...rest}>
+    <ContentBox>{children}</ContentBox>
   </BackgroundWrapper>
 );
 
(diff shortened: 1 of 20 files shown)


src/renderer/components/ViewContent/ViewContent.tsx
```diff
@@ -0,0 +1,11 @@
+import { BoxProps } from '@mui/material';
+import React from 'react';
+import { BackgroundWrapper, ContentBox } from './style';
+
+const ViewContent: React.FC<BoxProps> = (props) => (
+  <BackgroundWrapper {...props}>
+    <ContentBox>{props.children}</ContentBox>
+  </BackgroundWrapper>
```
karczewskiPiotr, line 8: @LostInStatic I have never considered this before but if the `props` are spread like that will it also spread the `children` to the `BackgroundWrapper` initially and then override it with the actual `children` passed in jsx? 🤔 

If we remove the `ContentBox` the `children` will still render to the `BackgroundWrapper`, albeit in a chaotic non styled way of course,  since `children` is just a prop in React. This just makes me wonder how React handles this internally and if there are any unnecessary assignments done under the hood because of the double passing of children. 

Can't find any immediate answers to this with a super quick search so perhaps it is just best to mitigate this by separating the `children` from other props:

```tsx
const ViewContent: React.FC<BoxProps> = ({ children, ...props }) => (
  <BackgroundWrapper {...props}>
    <ContentBox>{children}</ContentBox>
  </BackgroundWrapper>
);
```

LostInStatic: Yup, good point - my guess is that elements between the tags override the children passed as prop. If I try to set them explicitly ( `<BackgroundWrapper children={props.children} {...props}>` ), React protests about setting children twice. 

Anyway, I've applied your suggestion.

karczewskiPiotr (approved):

PR #147, opened 2021-12-01
App settings UI

### LINKED ISSUE #84: Implement app settings view

LostInStatic:
## Description

App settings view was styled according to the design. Some reusable components were modified so the will fit the design better (hopefully) when implementing other views. 

Root font was set at `10px`. Responsive fonts were removed from the theme, as they made exact implementation of the design difficult - hopefully to be re-added after brainstorming the general styling guidelines.

## Preview

image

## Linked issues

Resolves #84


karczewskiPiotr (changes requested):
In terms of the view itself there is only on thing that is bothering me and that is the lack of the top margin (20px in design) for the view title component that can be seen in the XD. ↓

image

We could also use the same style of the scrollbar as the one in the navigation bar (should be able to have a peek at the styles in `NavigationBar.scss`) for the view but that could perhaps be done for the whole app later on with separation for the main and alt themes.

image

LostInStatic:
@karczewskiPiotr Added padding. 

I've played around a bit with the scrollbar - I think it is best to leave it until we discuss the best solution. Styling the root element scrollbar proves to be difficult (especially with the switching theme), so the easiest solution would be having another component responsible for scrolling. My guesses are:
- `CustomRouter` (or additional wrapper component), but that requires moving the responsibility for providing the theme to it (so the scrollbar is rendered in scope of corresponding  `ThemeProvider`.) Theme to use could be specified in `RouterComponents.ts`. IMO 
- `ViewContent`, but that would require moving the responsibility for rendering subview to it (unless we want view and subview to have separate scrolling?)

karczewskiPiotr (approved):

karczewskiPiotr:
> @karczewskiPiotr Added padding.
> 
> I've played around a bit with the scrollbar - I think it is best to leave it until we discuss the best solution. Styling the root element scrollbar proves to be difficult (especially with the switching theme), so the easiest solution would be having another component responsible for scrolling. My guesses are:
> 
> * `CustomRouter` (or additional wrapper component), but that requires moving the responsibility for providing the theme to it (so the scrollbar is rendered in scope of corresponding  `ThemeProvider`.) Theme to use could be specified in `RouterComponents.ts`. IMO
> * `ViewContent`, but that would require moving the responsibility for rendering subview to it (unless we want view and subview to have separate scrolling?)

I'll note this down for the discussions as well then. 📓

PR #151, opened 2021-12-08
Issue128

### LINKED ISSUE #128: Create view for picking the project technologies

kubajaru:
## Description

Created Technology Picker according to AdobeXD

## Linked issues

Closes #128 

## Correct PR Checklist


src/renderer/views/TechnologyPicker/TechnologiesPicker.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import StyledSwitch from '../../components/Switch/Switch';
+import './TechnologiesPicker.scss';
+
+const TechnologiesPicker = () => {
+  const { t, i18n } = useTranslation();
+  let [useSCSS, changeUseSCSS] = useState(true);
+  let [useTP, changeTP] = useState(true);
```
karczewskiPiotr, line 9: We are not going to be using the `useState` in this component but for reference the definition should be `const` instead of `let`.

Instead of `useState` we're using redux and the reducer prepared in #144. So in order to get the information about the state of the sass and typescript toggles it is enough to use the `newPublication` selector and destructuring the `useSass` and `useTypescript` properties.

```ts
  const { useSass, useTypescript } = useSelector(newPublication);
```

These can then be passed to the switch as the `checked` props.

src/renderer/views/TechnologyPicker/TechnologiesPicker.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <p className='inline'> SCSS</p>
+        </div>
+        <div>
+          <StyledSwitch
+            size='small'
+            checked={useTP}
+            onChange={() => {
+              changeTP((useTP = !useTP));
+            }}
+          />
```
karczewskiPiotr, line 33: In the spirit of using redux the `onChange` handlers should be changed here as well. However, the use of the `changeUseSCCS` and `changeTP` was also wrong in itself. The function accepts a param. What you're doing here is assigning to `useSCSS` witht the `=` operator and then the result of that assignment is passed to the function. It's a weird thing in js that this works. The proper way to use the `useState`'s set function to change the value based on the previous one would be the follwing:

```ts
// The set function either accepts a value or a function that accepts the previous state value and returns a new one.
changeUseSCSS((prevUseSCCS) => !prevUseSCSS)
```

We're using redux for managing the state though so the `onClick` handlers should be changed to utilize that as well.
Therefore, you could define two handlers in the component as follows

```ts
// The properties selected from redux in the comment above
const { useSass, useTypescript } = useSelector(newPublication);
// We need to create a redux dispatch first

const toggleSass = () => {
  // We can then dispatch the setPublicationField action from addPublicationSlice to update the useSass
  dispatch(setPublicationField({ field: 'useSass', value: !useSass }));
};

const toggleTypescript = () => {
  // Same thing can be done for the useTypescript property
  dispatch(setPublicationField({ field: 'useTypescript', value: !useTypescript }));
};
```

These can then be passed to the switches

```tsx
<StyledSwitch
  size='small'
  checked={useSass}
  onChange={toggleSass}
/>
```

karczewskiPiotr (changes requested):

src/renderer/views/TechnologyPicker/TechnologiesPicker.scss
```diff
@@ -0,0 +1,30 @@
+.container {
```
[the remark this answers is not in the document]
kubajaru: Should I align the contents of the view to be in the center of the screen or leave them in the top right corner? 

Like that:
image

karczewskiPiotr (approved):

PR #153, opened 2021-12-10
Collaborators view

### LINKED ISSUE #129: Create view for adding collaborators

kmaciejewska:
## Description

There is a couple of issues that have been encountered during the implementation that need addressing and will be fixed accordingly. 
1. User avatars: for now they are set to display the first character of the username - not quite sure how the fetching of collaborators from github works? 
2. It seemed like some kind of validation was needed to prevent the user from adding just empty strings as role and username, so the styling of Button component has been extended to include 'disabled' state 
3. New component - CollabTable has been added. Also, CollabPicker's onAdd method has been slightly modified (couldn't come up with a better solution for now).

## Preview

1. 
image

2.
image

## Linked issues

closes #129 

## Correct PR Checklist

- [ ] I have assigned Reviewers
- [ ] I have assigned an Description label


src/renderer/components/CollabTable/CollabTable.tsx
```diff
@@ -0,0 +1,74 @@
...
+                <Styled.BorderedTC width='30px' />
+                <Styled.BorderedTC
+                  border
+                  align='center'
+                  width='30px'
+                  onClick={() => handleDelete(collaborator.id)}
+                >
+                  <Styled.Icon />
+                </Styled.BorderedTC>
+              </Styled.Row>
```
karczewskiPiotr, line 66: At the current stage the remove icon button is the only element on the page that is not focusable with the keyboard. It would be good to either give it a `tabindex` of `0` or wrap it with the `IconButton` component or example.

src/renderer/components/CollabPicker/CollabPicker.tsx
```diff
@@ -22,7 +22,7 @@ interface Props {
   selectPlaceholder?: string;
   textFieldPlaceholder?: string;
   onChange?: (value: Value | undefined) => void;
-  onAdd?: (value: Value) => void;
```
karczewskiPiotr, line 25: The changes in this component can be reverted I think.

karczewskiPiotr (changes requested):
The remove icon is not vertically aligned.

image

karczewskiPiotr (approved):

PR #155, opened 2021-12-15
Issue #126

### LINKED ISSUE #126: Create view with form for general project info

zucchero2304:
## Description

Added step one view in new project creation, validation of inputs and enabling/disabling "next" button depending on inputs

## Linked issues

closes #126

## Correct PR Checklist

- [x] I have linked an issue 


karczewskiPiotr:
@zucchero2304 make sure to resolve all the linting errors and I'll have a closer look after that. 🔍

src/renderer/views/AddProject/StepOne/StepOne.tsx
```diff
@@ -0,0 +1,80 @@
...
+import React, { useEffect, useState } from 'react';
+import ImagePicker from '../../../components/ImagePicker/ImagePicker';
+import TextField from '../../../components/TextField/TextField';
+import TextArea from '../../../components/TextArea/TextArea';
+import InputLabel from '../../../components/InputLabel/InputLabel';
+import './StepOne.scss';
+import { useTranslation } from 'react-i18next';
+
+export default function StepOne(props: any) {
+  const [imagePickerError, setImagePickerError] = useState(false);
```
karczewskiPiotr, line 11: I would default this to true for now so that it is possible to move on to other components. Could add a comment above it specifying why it was done.

src/renderer/views/AddProject/StepOne/StepOne.tsx
```diff
@@ -0,0 +1,80 @@
...
+          {t('StepOne.projectDescription')}
+        </InputLabel>
+        <TextArea
+          className='desc-input-field'
+          placeholder={t('StepOne.projectDescriptionDetails')}
+          error={descInputError}
+          onChange={(e) => handleDescValidation(e.target.value)}
+        />
+      </div>
+    </div>
```
karczewskiPiotr, line 78: Would be good to either scope the styles or transition to MUI `Box` components and use the `sx` prop since there is very little styling that is required for the divs.

src/renderer/views/AddProject/StepOne/StepOne.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </div>
+      <div className='right-column'>
+        <InputLabel error={nameInputError} typographyVariant='h3'>
+          {t('StepOne.projectName')}
+        </InputLabel>
+        <TextField
+          className='name-input-field'
+          placeholder={t('StepOne.projectNameDetails')}
+          error={nameInputError}
+          onChange={(e) => handleNameValidation(e.target.value)}
```
karczewskiPiotr, line 66: The validation could simply be done in `onBlur`. This removes the need for the triggered_ states.

karczewskiPiotr (changes requested):
There is no logic that sets the values for the project in redux. You can have a look at two other components for how to go about this. The gist is that the values in the inputs need to come from the `newPublication` selector and on change the `setPublication` action needs to be dispatched to set the correct fields in the store.

karczewskiPiotr (approved):

Issue #163, opened 2022-01-13
Convert all styling to MUI


karczewskiPiotr:
We aim to move all styling to use MUI. Sass should be removed from the project at the end of this procedure.

PR #166, opened 2022-02-05
M UI spacing

### LINKED ISSUE #161: Determine spacing options from XD design
It would be good to determine the spacing variants based on the XD design in order to add them to the MUI theme for use in the styling rework.

LostInStatic:
## Description

After analyzing the design I have come up with following spacing levels: 
- 1rem, mostly as text field padding
- 2.7rem, as a default distance between most elements 
- 5.5rem, between various things (more info below)
- 9rem, as a default distance between sections (such as control groups)

Button paddings were omitted - as they seem to be dependent on the inner font size, it would be best to implement them during typography rework. Logout view has been changed to showcase usage (other components will be converted when resolving #163).

The solution I chose is practical and IMO corresponds to the design well enough, however there are some potential issues:
- It's not really semantic. Alternative would be to extend the theme object with a custom property, but usage would be less straightforward and not permit shorthands described [here](https://mui.com/system/spacing/).
- The 5.5rem level is an arbitrary choice. Many elements in the design have between 48 and 60 px of spacing. (e.g. between button and text in `03 Project Info 1`, the label and the control in `10 New Project 3` and between the sections in subview. The most common value is 60px) I feel like it's a good proxy for all of the uses, but maybe a different value or more levels should be used.

## Linked issues

Resolves #161


karczewskiPiotr (approved):
> It's not really semantic. Alternative would be to extend the theme object with a custom property, but usage would be less straightforward and not permit shorthands described [here](https://mui.com/system/spacing/).

@LostInStatic I feel like it is hard to define spacing in such a way so that it makes sense. So I wouldn't bother with extending the theme in such a way.

> The 5.5rem level is an arbitrary choice. Many elements in the design have between 48 and 60 px of spacing. (e.g. between button and text in 03 Project Info 1, the label and the control in 10 New Project 3 and between the sections in subview. The most common value is 60px) I feel like it's a good proxy for all of the uses, but maybe a different value or more levels should be used.

Yeah, the values seem quite random so it would be hard to pick out specific levels. We can  stick to 5.5rem since the differences are not going to be too noticeable anyway.

PR #168, opened 2022-02-14
Styling conversion

### LINKED ISSUE #163: Convert all styling to MUI

LostInStatic:
## Description

All components have been converted to use styled components. Some components' code has been modified to improve semantics, reduce complexity or improve separation between logic and styling.

There is one component that has not been converted (`ProjectsList.tsx`) - is is a placeholder component with complex styling. IMO rewriting a placeholder version would be a waste of time, so SASS should be finally removed from dependencies when the component's actual implementation will be written.

## Linked issues

Resolves #163


karczewskiPiotr:
> There is one component that has not been converted (ProjectsList.tsx) - is is a placeholder component with complex styling. IMO rewriting a placeholder version would be a waste of time, so SASS should be finally removed from dependencies when the component's actual implementation will be written.

@LostInStatic this is totally fair and makes all the sense. 👌

karczewskiPiotr (commented):
@LostInStatic everything looks fine style wise. I'm getting some import errors though.

image

They go away if I import the `SxProps` from `@mui/system`. In that case we'd need to explicitly also add the `@mui/system` as a direct project dependency so that eslint doesn't complain.

FYI: The current installed version of MUI that I have is 5.0.4 so if you have a newer one (since the latest is **5.4.3**) and our `package.json` would allow it since we have it specified as `^5.0.4` then maybe they export the `SxProps` through the `@mui/material` in the new minor version. Let me know what version you have installed on your machine and I'll update if you're running a newer one. that supports this export.

Let me know and if the version would be a fix then I'll accept the PR.

---

There is also this for the `LangBox` styled component in `LoginComponent` but we are not doing and will not be doing SSR so it shouldn't really matter at all.

image

LostInStatic:
@karczewskiPiotr 
> The current installed version of MUI that I have is 5.0.4 so if you have a newer one (since the latest is 5.4.3) and our package.json would allow it since we have it specified as ^5.0.4 then maybe they export the SxProps through the @mui/material in the new minor version. Let me know what version you have installed on your machine and I'll update if >you're running a newer one. that supports this export.

Yup, that was the issue. Bumped the version up, seems to not break anything.

> There is also this for the LangBox styled component in LoginComponent but we are not doing and will not be doing SSR so it shouldn't really matter at all.

I fixed it anyway, as ~2 mins of work seems like a fair price for seeing less red in the console 😄

karczewskiPiotr (approved):

Issue #170, opened 2022-02-15
Implement form validation for AddProject view


karczewskiPiotr:
The `description`  of the project should definitely be an optional field. 😅

We can use [Formik](https://github.com/jaredpalmer/formik) and [Yup](https://github.com/jquense/yup) for the validation.

PR #172, opened 2022-03-04
Refactor AddProject view 

### LINKED ISSUE #169: Refactor AddProject view
The `AddProject` view needs refactoring. Things that jump out immediately:
 - [x] Sub-components are all over the place. Some in `subcomponents`, some in root folder of `AddProject`, some in higher `components` folder.
 - [x] Naming of the subcomponennts is a bit off. It should be more semantic.
 - [x] The `renderStepComponent` function has to many cases.
 - [x] Perhaps it would also make sense to memo the rendered step component instead of calling the function at each render.

LostInStatic:
## Description
The main component was cleaned-up. Subcomponents were moved to the `subcomponents` folder and renamed appropriately. Step controls were moved to a separate subcomponent, missing features were partially implemented.  

The negation of  `nextButtonEnabled` is unnecessarily confusing - I will clean it up when dealing with #170  

## Linked issues

Closes #169


karczewskiPiotr (changes requested):
@LostInStatic good call on the next button handling. Refactoring is great. Just a couple things I noticed on the views themselves that could be tackled here:


- [ ] On many occasions the `x` button in collaborators table is having it's right border cut off. ↓

    image

    If the reason for this will turn out to be hard to pinpoint and therefore fix you can make it into an issue. Not worth wasting time on now if it would be a pain to deal with.

---

- [ ] The header of the step should be uniform across all steps but on step 3 it is `bold` instead of `regular` like in the XD design.

    image

---

- [ ] In the XD design steps 2-4 have a view header that also includes the name of the project that was input by the user in step 1 so we could add that as well. ↓

    image

LostInStatic:
@karczewskiPiotr  your suggestions were implemented. I cleaned up some typography I must've missed earlier.

The way I did it, the header is updated real-time on step one. It was the most straightforward implementation and I find the effect neat, but if we deem it too distracting, it can be removed easily by adding `step > 1` as a second condition.

karczewskiPiotr (approved):
Just removed some redundant brackets not to bother you with it.

PR #173, opened 2022-03-12
AddProject form validation

### LINKED ISSUE #170: Implement form validation for AddProject view

LostInStatic:
## Description
Yup and Formik had been added to dependencies. 

The AddProject validation is somewhat a bare-minimum one. Only the project name is properly validated - all other fields in other steps are either optional, not implemented yet or do not require validation by nature (switches). 

The implementation is a bit awkward due to the way the component relays it's state to redux constantly - a function (inside `useEffect`) handles updating the redux state every time a field is changed and the validation passes. We should discuss if it is the optimal solution and rewrite it if that's not the case..

I fixed the translation naming while at it.

## Linked issues

closes #170

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "dotenv": "^8.2.0",
     "electron-redux": "^1.5.4",
     "electron-squirrel-startup": "^1.0.0",
+    "formik": "^2.2.9",
     "i18next": "^21.6.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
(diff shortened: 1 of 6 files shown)


karczewskiPiotr (changes requested):
@LostInStatic I'd say we can stick with the approach of the `useEffect` as otherwise we'd need to somehow reverse the responsibility for the next button press down to the steps. However with these steps being quite simple we can totally leave it like this.

karczewskiPiotr (approved):

PR #174, opened 2022-03-12
Package manager view

### LINKED ISSUE #127: Create view for choosing package manager

LostInStatic:
## Description

Package manager picker was implemented according to the design and comments. 

## Linked issues

Closes #127

### DIFF
--- a/src/renderer/views/AddProject/subcomponents/PackageManagerPicker/PackageManagerPicker.tsx
+++ b/src/renderer/views/AddProject/subcomponents/PackageManagerPicker/PackageManagerPicker.tsx
@@ -14,7 +14,8 @@ const PackageManagerPicker = () => {
   const { packageManager } = useSelector(newPublication);
   const dispatch = useDispatch();
 
-  const setPackageManager = (value: NewPublication['packageManager']) => {
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.value as NewPublication['packageManager'];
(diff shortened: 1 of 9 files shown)


src/renderer/views/AddProject/subcomponents/PackageManagerPicker/PackageManagerPicker.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+      <Box mb={2}>
+        <RadioGroup
+          defaultValue={packageManager}
+          onChange={(e) =>
+            setPackageManager(
+              e.target.value as NewPublication['packageManager']
+            )
+          }
+        >
```
karczewskiPiotr, line 35: @LostInStatic I think it might just be cleaner to have a handler function instead of defining the `onChange` inline. Even though for now we don't even allow a choice here. 😅

```suggestion
  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    const value = event.target.value as NewPublication['packageManager'];
    dispatch(setPublicationField({ field: 'packageManager', value }));
  };

  return (
    <Box>
      <Typography variant='subtitle1' component='p' mb={3}>
        {t('AddProject.PackageManager.message')}
      </Typography>

      <Box mb={2}>
        <RadioGroup defaultValue={packageManager} onChange={handleChange}>
```

karczewskiPiotr (changes requested):

PR #199, opened 2022-04-15
Notification spawning system

### LINKED ISSUE #198: Create the notification spawning system

karczewskiPiotr:
## Description

Notifications can be created by dispatching a `sendNotification` action and removed by dispatching the `dismissNotification` action. They are stored in redux so that there is a possibility to dispatch them from the back-end as well. This constrains the content of the notification a little bit due to redux data being serializable so we can not just send a react node (_although one could use react dom to render a complex component to string with the loss of interactivity_) to be saved in state like it would be possible with context. 

The way to spawn a notification:

```js
dispatch(sendNotification({
  title: "Title",
  message: "message",
  type: 'success',
  autoDismiss: true, // if not set the default is false
  delay: 2000, // if not set the default is 4000 | only applicable if autDismiss is true
})
```

The way to dismiss a notification:

```js
dispatch(dismissNotification(id));
```

## Preview

image

notification

## Linked issues

Closes #198

### DIFF
--- a/src/renderer/components/CloseButton/CloseButton.tsx
+++ b/src/renderer/components/CloseButton/CloseButton.tsx
@@ -1,5 +1,5 @@
 import { IconButton } from '@mui/material';
-import ClearIcon from '@mui/icons-material/Clear';
+import { Close } from '@mui/icons-material';
 import { t } from 'i18next';
 import React from 'react';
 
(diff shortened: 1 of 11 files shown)


LostInStatic (approved):
Looks good to me. 

There was a reusable `CloseButton` component that could be used instead of assembling the button manually in the `Notification.tsx`. I liked your implementation better though, so I also updated the reusable component.

The slice name was wrong.

Since both changes were simple, I've just pushed two commits to the branch.

PR #200, opened 2022-04-15
Fix filesystem errors

### LINKED ISSUE #196: Fix filesystem actions failures
## Description

When performing file actions with incorrect path, the application throws an error on electron's main process. This makes the interface only display an error message and prevents the user from correcting the mistake.  I've noticed two handlers causing this behavior:

- `settings:read`: When the `publab-settings.json` is missing from PubLab's appdata folder or the whole folder is not yet generated ; seems like the read should just be omitted if the file is not found.
- `publications:find-local`: When the default folder path is wrong. It is possible to set the wrong path during the first time setup by picking a folder and then changing the path manually ; probably the input should be validated before the state update as in `AppSettings` view.

IMO the handler errors should be displayed in a way that does not unnecessarily interfere with the application functionality as often as possible.  The root cause could be just a bad input or stale data, easily correctable by the user. However, simply updating the logic/validation will address the currently occurring problems - the holistic approach might as well be discussed later.

karczewskiPiotr:
~~❗ ❗ This branch has been created from the #199. Therefore this PR will be a draft till it is merged. The changes should not be looked at before #199 is merged as it will show both.~~

## Description

- The app no longer crashes if the settings file is not present.
- The directory picker in the first time view is validated and shows an error message on blur of the field

## Preview

filesystem

## Linked issues

Fixes #196


LostInStatic (dismissed):
Looks good to me.

PR #208, opened 2022-04-29
Improve PublicationModification type

LostInStatic:
## Description

Instead of a statically written union type, it is possible for it to be generated based on `Publication` keys and associated types.

### DIFF
--- a/src/shared/redux/slices/loadPublicationsSlice.ts
+++ b/src/shared/redux/slices/loadPublicationsSlice.ts
@@ -5,29 +5,17 @@ import { RootState } from '../rootReducer';
 // type created to avoid an issue with both unclear modification type
 // and index access writes on union of keys, which would complicate the
 // code significantly more
-type PublicationModification =
-  | {
+
+type PublicationModification = Exclude<
+  {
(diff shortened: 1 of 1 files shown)


src/shared/redux/slices/loadPublicationsSlice.ts
```diff
@@ -5,29 +5,17 @@ import { RootState } from '../rootReducer';
...
-  | {
+
+type PublicationModification = Exclude<
+  {
+    [FieldName in keyof Publication]: {
       id: string;
-      field: keyof Pick<Publication, 'status'>;
-      value: 'cloned' | 'remote';
+      field: FieldName;
+      value: Publication[FieldName];
```
karczewskiPiotr, line 14: If we'd want the array values to be the type of the array element we could do the following as well. But I don't see that really coming in handy.

```ts
value: Publication[FieldName] extends Array<any>
        ? Publication[FieldName][0]
        : Publication[FieldName];
```

The most versatile thing for array properties would be having `add`, `remove` and `set` actions but you can see ass you go if they are needed.

karczewskiPiotr (approved):

PR #213, opened 2022-05-10
Project settings view

### LINKED ISSUE #197: Create the project settings view
## Description

The `Project settings` view should be implemented according to the design.

LostInStatic:
## Description

The view is implemented with the intended functionality. There are slight differences between this implementation and the design - I tried to keep it consistent between the components and avoid unnecessary complications at this point.

The deletions do not trigger confirmation dialog. It seems unnecessary (as saving the changes requires clicking another button), but I might be mistaken on that part.

## Linked issues

 closes #197


src/renderer/views/Settings/subcomponents/SnippetsManager/SnippetsManager.tsx
```diff
@@ -0,0 +1,92 @@
...
+                      ...snippets.slice(0, index),
+                      ...snippets.slice(index + 1),
+                    ],
+                  })
+                }
+              />
+            </Styled.SnippetHeader>
+
+            <Styled.SnippetName
+              value={snippets[index].name}
```
karczewskiPiotr, line 56: @LostInStatic any reason for not doing `snippet.name` here and for the `code` down below?

karczewskiPiotr:
@LostInStatic I have just fixed some leftover typing issues on the back-end that I noticed during the review but other than that looks great. 🚀

src/renderer/views/Settings/subcomponents/SnippetsManager/SnippetsManager.tsx
LostInStatic: I guess I was in the `snippet[index]` mood? Fixed.

karczewskiPiotr (approved):

PR #215, opened 2022-05-19
206 create files view

### LINKED ISSUE #206: Create files view

LostInStatic:
## Description

The view was implemented. The file statuses are random until git integration will be finished.

Tree nodes can be interacted with by double clicking or pressing `enter` (default TreeView behavior overriden). It will move the root path if it is a directory or open it in the default OS application if it is a file. 

The "type" column was omitted, as it would not display new information (that can't be conveyed by the filename/icon). AFAIK, anything more useful than getting the extension from the filename (an maybe mapping it to a human-readable string) requires including another dependency / writing our own utility. 

## Linked issues

Closes #206


src/renderer/components/FileDisplay/FileDisplay.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <FolderIcon htmlColor={statusColor} />
+        ) : (
+          <FileIcon htmlColor={statusColor} />
+        )}{' '}
+        {entry.name}
+      </Styled.DataField>
+      <Styled.DataField sx={{ width: widths[1] }}>
+        <Typography variant='body2'>
+          {`${dateM.getDay()}/${dateM.getMonth()}/${dateM.getFullYear()}`}
+        </Typography>
```
karczewskiPiotr, line 44: I believe we should use the same date format via util as in the `ProjectRow`.

```suggestion
        <Typography variant='body2'>
          {getDateString(
            entry.details.dateModifiedMs,
            i18n.language as SupportedLangCode
          )}
        </Typography>
```

Suggestion would be missing imports I guess so better to manually edit this.

src/renderer/components/FileDisplay/FileDisplay.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const [statuses, colors] = [
+  ['added', 'modified', 'deleted', 'moved', 'unchanged'],
+  ['green', 'blue', 'red', 'orange', 'primary'],
+] as const; // placeholder
+
+const FileDisplay = ({ entry, treeLevel }: Props) => {
+  const { t } = useTranslation();
+  const dateM = new Date(entry.details.dateModifiedMs);
+  const statusIndex = useMemo(() => Math.floor(Math.random() * 5), []);
```
karczewskiPiotr, line 23: Sometimes the placeholder status is not shown in the cell but I suppose that's because of this randomization going out of bound of the array maybe?

src/shared/types/api.ts
```diff
@@ -5,6 +5,23 @@ export type IpcEventHandler<R = any> = (
...
+  };
+}
+
+export interface FolderInfo extends DirectoryEntryBase {
+  directory: { isDirectory: true; content: DirectoryEntryInfo[] | undefined };
+}
+
+export interface FileInfo extends DirectoryEntryBase {
+  directory: { isDirectory: false; content: null };
+}
```
karczewskiPiotr, line 21: Any reason for the missing content to once be `null` and once `undefined`?

karczewskiPiotr (changes requested):
Found a bug unfortunately so for a little context to the gif:

1. The first time I go into the Files view and go out of `publication` by pressing the dots once it all works well.
2. Then I go back in to the `publication` and go back via dots by pressing twice because I somehow did that for no real good reason and it still works well.
3. If I reload the app or just start it clean without interacting with Files view and attempt to go back out of `publication` via double click of dots the app crashes. 

files

It seems like there is a small amount of time during, which the `ToParentFolder` disappears but the old folder contents are still displayed so the second press tries to open some folder that appears under the mouse because the table shifted. The layout shift can be seen below. Maybe we could leave the first row visible at all times but disabled when the user can not go back anymore.

files-2

src/main/api/files/readDirectory.ts
```diff
@@ -0,0 +1,60 @@
...
+const readDirectory: IpcEventHandler<DirectoryEntryInfo[]> = async (
+  _,
+  dirPath: string,
+  options?: { depth?: number; withDetails?: boolean }
+) => {
+  const optionsWithDefaults = {
+    depth: 0,
+    withDetails: false,
+    ...options,
+  };
```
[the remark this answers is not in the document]
LostInStatic: Sure, corrected.

src/main/api/files/readDirectory.ts
```diff
@@ -0,0 +1,60 @@
...
+        ? { dateModifiedMs: stats.mtimeMs }
+        : undefined;
+
+      return {
+        name: dirent.name,
+        directory,
+        details,
+      };
+    })
+  );
```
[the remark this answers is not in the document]
LostInStatic: Yeah, that works, seems obvious in hindsight :D

src/renderer/components/FileDisplay/FileDisplay.tsx
LostInStatic: Must've written it without thinking while doing the first draft of the component, corrected.

src/renderer/components/FileDisplay/FileDisplay.tsx
LostInStatic: There was no logic bug, but empty strings in translations - I could not find a good word for 'unchanged', so did an experiment with just leaving them blank and forgot to do something with it (e.g. a conditional in the component instead of an empty translation) before doing the PR. Filled them in for now.

src/shared/types/api.ts
LostInStatic: My reasoning was, that `null` means explicit lack of value (as non-directory cant have content in this case) and `undefined` means lack of value for unspecified reasons (in this case, value not yet assigned as data is not loaded yet), but maybe it's too confusing? My first implementation had simply `{ isDirectory: false }` for files, but having the same shape for both file and folder was simpler to deal with.

LostInStatic: I guess it could be simplified to just being `{ isDirectory: true; content: DirectoryEntryInfo[] | null }` for both file and folder, as: 
 - If `isDirectory` is false, it is a file and explicitely has no content
 - If `isDirectory` is true and `content` is null, it is a directory and content (explicitely) has not been loaded yet
 - If the directory is empty, `content` should be an empty array

What do you think?

LostInStatic:
> Found a bug unfortunately so for a little context to the gif:

I've corrected the conditionals to prevent layout shifts. I've also made the interaction consistent with other interactions (only on `enter` press or double click) in a separate commit - it seems to be better UX for me, but it makes navigation slightly more tedious.

src/shared/types/api.ts
karczewskiPiotr: @LostInStatic I like this approach. Empty array for empty directory is also more controllable since you will not need to perform checks before iterating on the array. In terms of null vs undefined it's always a matter of preference due the behaviors of both . Some will only use null and some only undefined. I'd say just pick one for this type. Using both null and undefined in the same type could get confusing and cause weird conditional checks sometimes.

karczewskiPiotr:
> I've corrected the conditionals to prevent layout shifts. I've also made the interaction consistent with other interactions (only on `enter` press or double click) in a separate commit - it seems to be better UX for me, but it makes navigation slightly more tedious.

@LostInStatic Nice! I also added the side to side layout shift prevention when navigating between small/big directories. The double click is a bit more tedious but I think we can leave it like that. It might not even be used that much since folders are collapsible.

karczewskiPiotr (approved):

PR #217, opened 2022-06-06
Implement commit view functionality

### LINKED ISSUE #214: Create the commits view

### LINKED ISSUE #216: isomorphic-git: TypeError: ign.test is not a function
## Description

When using `isomorphic-git`'s status related functions, e.g. `statusMatrix`, the "ign.test is not a function" error is thrown. Ign refers to the object derrived from the [`ignore`](https://www.npmjs.com/package/ignore) package. 

Apparently, the `.test` method was added in version 5.0.0. In theory, `isomorphic-git` correctly references its own version of the dependency (5.2.0).
image

However, I was able to stop the error from happening by adding 
```
 "resolutions": {
    "ignore": "5.2.0"
}
```
to `package.json` and doing a `yarn install`. 

I am a bit stumped at this point. Maybe the way the application is built somehow mixes up different dependency versions?

Anyway, the issue might not need other resolution than what I already did in [this commit](https://github.com/vnLab-Lodz/PubLab/commit/011c2e05a3cc85c872a8580f58b86154583170db), but I decided to document it in case this solution causes other problems down the line.

LostInStatic:
## Description

The view is implemented mostly according to the design - the "history" tab and tab controls were omitted, files are grouped by folder instead of just listing them. 

The files view was refactored to utilize the newly created slice instead of gathering its own data. 

Placeholder refresh button is still there - will be removed once an autorefresh method will be implemented.

Bugs I've stumbled upon are described in #216 and #218. As we deemed the `package.json` `resolutions` fix acceptable after small improvements, this PR will close #216.

## Linked issues

Closes #214, #216


src/renderer/components/FileDisplay/FileDisplay.tsx
```diff
@@ -32,24 +36,25 @@ const FileDisplay = ({ entry, treeLevel }: Props) => {
...
-            entry.details.dateModifiedMs,
-            i18n.language as SupportedLangCode
-          )}
+          {item.stats &&
+            getDateString(
+              item.stats.mtimeSeconds * 1000,
+              i18n.language as SupportedLangCode
+            )}
         </Typography>
       </Styled.DataField>
```
karczewskiPiotr, line 54: @LostInStatic The `DataField` needs to be in the tree even if the `item.stats` are `undefined` so that the layout is preserved?

src/renderer/views/Changes/subcomponents/CommitForm/CommitForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+  });
+  const tree = useSelector(selectRepoTree);
+  const commited = tree
+    ? repoTree.search(
+        tree,
+        (node) =>
+          checkStatus.isChanged(node.status) &&
+          checkStatus.isStaged(node.status)
+      )
+    : [];
```
karczewskiPiotr, line 54: @LostInStatic do you think this would potentially benefit from a `useMemo`?

karczewskiPiotr (commented):
@LostInStatic overall the functionality seems to work quite well. Couple caveats I can think of:

- Changes are nicely commited but not pushed to the remote. I don't recall any place in the app for that to happen so probably good for it to be implicit when the changes are commited. Other idea would be for an icon button next to the refresh one.
- The "choose all" option takes quite a bit of time for lots of files so perhaps we could spawn a loader for the duration of that with the `addLoader` action.
- It might be worth to add "remove all" option as well next to the "choose all".

src/renderer/views/Changes/subcomponents/CommitForm/CommitForm.tsx
LostInStatic: IMO not really - there are two cases when the `commited` array could not be recomputed (`closeForm` prop is ignored, it should not change without parent rerendering):
 - when the parent component rerenders: It would require a different setup than just wrapping it in `useMemo`. Also, for now in all the cases (I can think of) the component should be unmounted anyway.
 - when `tree` is updated, but stays the same: AFAIK it would require a deep comparison, which would be as resource-intensive as the recomputation.

src/renderer/components/FileDisplay/FileDisplay.tsx
LostInStatic: Yes. It happens when the file is deleted from the workdir. We could display a message ("file removed" or something like that) if the reason the field is empty might be unclear for the user.

package.json
```diff
@@ -153,6 +153,6 @@
     "yup": "^0.32.11"
   },
   "resolutions": {
-    "ignore": "5.2.0"
+    "isomorphic-git/ignore": "^5.2.0"
```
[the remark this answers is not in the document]
LostInStatic: Yeah - as I told you before, for some reason only switching to a freshly created publication caused the error for me, but reverting the resulutions to just `ignore` did the trick. Weird.

src/renderer/components/FileDisplay/FileDisplay.tsx
karczewskiPiotr: Should be fine, can always add it later. 👍

LostInStatic:
@karczewskiPiotr 
> @LostInStatic is the push to repo and deselect all functionality going to be added to this PR or am I free to approve it otherwise?

I'm going to add them - I got sidetracked while working on the select/deselect all, as I realized that updating really should be optimised (by allowing partial refresh) to avoid annoying loading times on bigger publication repos. Mostly done with that now, should be finished tommorow.

LostInStatic:
Ready for review. 

The `push` function will push to a different branch (unless explicitly told otherwise) when the app is launched in development mode, mostly because I can see myself just opening a random publication and testing something without thinking about where it'll end up 🙈 

The status refresh button ended up around the title - seems like a logical spot, but I'm not 100% convinced about the placement and the looks.

PR #220, opened 2022-07-08
Implement project list search

### LINKED ISSUE #183: Implement project list search

LostInStatic:
## Description

The functionality was implemented, the UI is based on the search box in the Files view design. 

It is possible to search by tag or by title. Right now only the publications that match all search terms are shown, but it will be easy to change it to "match any term" or tie the behavior to a switch/checkbox if we decide it will be an improvement.  

The layout is somewhat resistant to excessive input, but more than 2-3 lines will look off. 

## Linked issues

Closes #183

### DIFF
--- a/src/renderer/views/ProjectsList/subcomponents/ProjectSearch/ProjectSearch.tsx
+++ b/src/renderer/views/ProjectsList/subcomponents/ProjectSearch/ProjectSearch.tsx
@@ -1,0 +1,30 @@
+import { Autocomplete } from '@mui/material';
+import React from 'react';
+import { Publication } from '../../../../../shared/types';
+import SearchField from '../../../../components/SearchField/SearchField';
+
+interface Props {
+  publications: Publication[];
+  onChange: (terms: string[]) => void;
+}
+
+const ProjectSearch: React.FC<Props> = ({ publications, onChange }) => (
+  <Autocomplete
+    options={getAutocompleteOptions(publications)}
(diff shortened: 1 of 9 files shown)


src/renderer/views/ProjectsList/subcomponents/ProjectSearch/ProjectSearch.tsx
```diff
@@ -0,0 +1,42 @@
...
+    />
+  );
+};
+
+export default ProjectSearch;
+
+function getAutocompleteOptions(publications: Publication[]) {
+  const options = [] as string[];
+  publications.forEach(({ tags }) => {
+    tags.forEach((tag) => !options.includes(tag) && options.push(`#${tag}`));
```
karczewskiPiotr, line 39: @LostInStatic this will add all tags as autocomplete options and in result provide items with same keys for React to complain about. This will be due to the fact that `options` array includes items in the format of `#tag_1` and the `tag` itself will be `tag_1` . Therefore, `includes` condition will always throw `false` for duplicate keys.

karczewskiPiotr (changes requested):

src/renderer/views/ProjectsList/subcomponents/ProjectSearch/ProjectSearch.tsx
LostInStatic: Oops! Amended.

karczewskiPiotr (approved):

PR #227, opened 2022-07-26
Fix login screen styles

karczewskiPiotr:
## Description

- The view now fits within the viewport.
- Some additional margins have been given to elements to protect from being squashed on tiny wee viewports.

## Preview

image


LostInStatic (approved):
It does fit, LGTM :rocket:

PR #228, opened 2022-07-26
Gatsby-config modification

### LINKED ISSUE #223: Settings view should also modify gatsby-config in the publication
The settings change should also modify the data in the `gatsby-config.ts` file via the `gatsbyConfigHandler`. This can be done in a similair fashion to what happens in the `generate` function here ↓

 https://github.com/vnLab-Lodz/PubLab/blob/f2040bc06dae0a08e6568ad022eec4c4ecd39ce2/src/main/api/publications/generate.ts#L50-L55

LostInStatic:
## Description

Now config changes are also reflected in `gatsby-config`. Additionaly, publication `package.json` is updated too.

Changed files will be immediately commited and pushed to the remote. 

## Preview

electron_MIrwOLbMh2

## Linked issues

e.g. closes #223

### DIFF
--- a/src/main/api/publications/updateConfig.ts
+++ b/src/main/api/publications/updateConfig.ts
@@ -1,7 +1,11 @@
+/* eslint-disable no-param-reassign */
+import path from 'path';
 import { IpcEventHandler } from 'src/shared/types/api';
 import createConfigFileHandler, {
   Config,
 } from '../../lib/configurationFileHandler';
+import createGatsbyConfigHandler from '../../lib/gatsbyConfigHandler';
+import createPackageHandler from '../../lib/packageHandler';
 
 const updateConfig: IpcEventHandler = async (
   _,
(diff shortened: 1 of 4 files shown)


src/renderer/views/Settings/Settings.tsx
```diff
@@ -30,70 +32,78 @@ const Settings = () => {
...
+            color='green'
+            sx={{ m: 0 }}
+            isMajor
+            fullWidth
+            disabled={!canSubmit}
+            onClick={async () => {
+              const id = uuidv4();
+              setLoaderID(id);
+              await updateConfig(project.dirPath, changes);
+              await gitCommit('Config update\n\n[publAB automatic commit]');
```
karczewskiPiotr, line 88: @LostInStatic some casing issues creeped in to the commit message

src/main/api/publications/updateConfig.ts
```diff
@@ -11,6 +19,48 @@ const updateConfig: IpcEventHandler = async (
   const configHandler = createConfigFileHandler({ dirPath });
   const oldConfig = await configHandler.getConfig();
   await configHandler.setConfig({ ...oldConfig, ...changes });
+  await stageChanges(CONFIG_NAME, dirPath);
+
+  if (changes.name || changes.description) {
```
karczewskiPiotr, line 24: @LostInStatic you can usually just return early with inverse condition instead of wrapping the code in an `if` in these situations.

karczewskiPiotr (changes requested):

src/main/api/publications/updateConfig.ts
LostInStatic: Sure, amended.

src/renderer/views/Settings/Settings.tsx
LostInStatic: Funny, I was SURE this is how we style it at that moment

karczewskiPiotr (approved):

PR #231, opened 2022-07-31
Guard repo fetch if token is not present

### LINKED ISSUE #229: Changing language on login screen causes error in publications:find-remote
## Description

electron_TNWoV9YREc

After closing the error prompt app seems to function normally, so it is not crucial to fix it right now.

## Error message

```
Error invoking remote method 'publications:find-remote': HttpError: Requires authentication
Call Stack
 EventEmitter.i.send.i.invoke
  js2c/renderer_init.js:71:535
```

## Steps to replicate

1. Get to the login screen
2. Change the language

karczewskiPiotr:
## Description

Guard to not fetch remote publications when the token is missing has been put in place.

## Linked issues

Fixes #229


LostInStatic (approved):
LGTM, seem to work fine.

PR #232, opened 2022-08-15
Fix authorization in the packaged app

### LINKED ISSUE #230: Make auth functional in packaged app

karczewskiPiotr:
## Description

⚠️ In order for the auth to work the callback url on github needs to be set to `publab://auth/callback`.

- Authorization in the packaged app is now working, 
- Environmental variables are properly loaded in the packaged app.

## Linked issues

Closes #230

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,25 +1,33 @@
+import path from 'path';
 import { app, BrowserWindow } from 'electron';
-import { configStore } from '../shared/redux/configureStore';
-import registerApiHandlers from './api';
 import installDevToolsExtensions from './devToolsExtensions';
+import registerApiHandlers from './api';
+import { configStore } from '../shared/redux/configureStore';
 
 declare const MAIN_WINDOW_WEBPACK_ENTRY: any;
 declare const MAIN_WINDOW_PRELOAD_WEBPACK_ENTRY: any;
 
-require('dotenv').config();
-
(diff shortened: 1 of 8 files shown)


src/main/index.ts
```diff
@@ -1,25 +1,33 @@
...
+let mainWindow: InstanceType<typeof BrowserWindow>;
+const gotTheLock = app.requestSingleInstanceLock();
 
-export const mainStore = configStore('main');
+// Handle creating/removing shortcuts on Windows when installing/uninstalling.
+if (require('electron-squirrel-startup')) app.quit();
 
-registerApiHandlers();
+if (process.defaultApp) {
+  if (process.argv.length >= 2) {
```
LostInStatic, line 19: This bit confuses me - I don't doubt the correctness, but I am not sure what is the case it covers. Could you add a comment?

LostInStatic, line 19: This bit confuses me - I don't doubt the correctness, but I am not sure what is the case it covers. Could you add a comment?

LostInStatic (changes requested):

src/main/index.ts
karczewskiPiotr: @LostInStatic added a bit of an explanation and link to docs but it is not exactly clear what passing the current js path does for the protocol set-up. I assume it's for opening links when app is running so it point to the currently running one instead of setting up a new one or something like that.

LostInStatic (approved):

PR #239, opened 2022-08-21
222 project cover image

### LINKED ISSUE #222: Implement the project cover image selection and saving
I think the creator step can display the image from the users disk and on creation the image should be copied over to the publication repository (`static/images`) from which it can be displayed in other places of the app.

LostInStatic:
## Description

Now cover image can be picked. It will be copied to the publication directory as `cover.***`. Image path (relative) is saved to `publab.config.json` to keep track of the extensions.

## Preview

PubLab_dlTwWVAmFf

## Linked issues

Closes #222


karczewskiPiotr (changes requested):
@LostInStatic so in the end we don't have an asset slice but rather a hook that creates a uri for the given asset path on the machine, right? Sounds reasonable. 👍

karczewskiPiotr (approved):

LostInStatic:
@karczewskiPiotr Added two things previously missed: 
- cleanup after changing the cover picture, as the file won't be overwritten if extensions don't match
- button for removing the picture
electron_2Uqg6f9mda

karczewskiPiotr (approved):
@LostInStatic ah that's a good update, great work! Remove could definitely come in handy.

PR #240, opened 2022-08-25
Collaborator addition - username verification

### LINKED ISSUE #221: Implement the proper collaborator addition and removal from the project
We are not aiming for auto complete but rather a check if the GitHub user exists when the add attempt is performed. If the user does not exist the error notification can be shown.

LostInStatic:
## Description

Collaborator username will be verified through the GitHub API and rejected if user does not exist or is already added. Prompt will be displayed below when the username is wrong.

## Preview

electron_KgqALGP9OK

## Linked issues

Closes #221

### DIFF
--- a/src/main/api/gitHub/getPublicUserData.ts
+++ b/src/main/api/gitHub/getPublicUserData.ts
@@ -1,0 +1,21 @@
+import { Octokit } from '@octokit/rest';
+import { mainStore as store } from 'src/main';
+import { IpcEventHandler } from 'src/shared/types/api';
+import { createLogger } from '../../logger';
+
+const getPublicUserData: IpcEventHandler = async (_, username: string) => {
+  try {
+    const token = store.getState().currentUser.auth.accessToken?.value;
+    const octokit = new Octokit({ auth: token });
+    const userData = await octokit.rest.users.getByUsername({ username });
+    return userData;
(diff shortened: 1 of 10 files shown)


karczewskiPiotr (approved):
@LostInStatic seems to work nicely in both new project and project settings views. 👍 

I am not super convinced that the control over the state of the `CollabPicker` needed to to be outsourced to a parent component. It feels like there was a potential for addition of a `preventDefault` param to the `CollabPicker.onAdd` prop alongside making it async, which would allow to stop the clearing of the fields. Therefore, it would be possible to await for that callback that could take care of everything instead of using the `usePromiseSubscription` + `useEffect` + `useState` combination. 

The benefit of your approach is that you get the errors and pending statuses by default from that hook and don't need to manage it with the downside of having to handle part of the addition process in the additional `useEffect` and having that `useState` to invalidate the query.

Syntax might be a bit harder to wrap your head around with but since it all works I don't think it's worth changing anything so I'm happily approving this PR.

LostInStatic:
@karczewskiPiotr 
Yeah, it is a bit awkward - initially I wanted to change as little of the component logic as possible, but then I realised, that the input state was kept in both components and (kind of) synchronised. Making `CollabPicker` a controlled component was the simplest solution.

PR #246, opened 2022-09-09
Fix first time view logic

### LINKED ISSUE #237: Fix the first time view logic

karczewskiPiotr:
## Description

- The first time view will only appear if the directory path is missing.
- Local storage flag is not cleared on log out.

## Linked issues

Closes #237

### DIFF
--- a/src/main/git/gitAuthorization.ts
+++ b/src/main/git/gitAuthorization.ts
@@ -92,5 +92,5 @@ export async function requestAccessToken(code: string): Promise<any> {
 }
 
 export const terminateSession = async (): Promise<void> => {
-  await session.defaultSession.clearStorageData();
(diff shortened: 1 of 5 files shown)


LostInStatic (approved):
LGTM. The only caveat is that the first time view will not be displayed if the `defaultDirPath` property is somehow missing from `publab-settings.json`. IMO no reason to fix that - chances of that happening accidentally are miniscule at best and the user would still be able to fix it through App Settings view.

karczewskiPiotr:
> LGTM. The only caveat is that the first time view will not be displayed if the `defaultDirPath` property is somehow missing from `publab-settings.json`. IMO no reason to fix that - chances of that happening accidentally are minuscule at best and the user would still be able to fix it through App Settings view.

Was a quick fix so added it.

PR #248, opened 2022-09-12
Support for polish letters in `publab.config.json`

### LINKED ISSUE #247: Decoding `publab.config.json` with polish letters fails

karczewskiPiotr:
## Description

With UTF-8 we support polish letter in the config file. Previosuly this resulted in https://github.com/vnLab-Lodz/odszkolnic_web failing to register as a remote publication in the app.

@LostInStatic not much to review here but would be grand if you could check if "Odszkolnić! Deschool!" shows up in your publications list now.

## Linked issues

Closes #247


LostInStatic (approved):
Works good on my end, no problems with listing nor cloning that repo.

Issue #249, opened 2022-09-15
Handling cover image saves unwanted metadata to `publab.config.json`


karczewskiPiotr:
Adding an image from settings pushes more than just the `imagePath` to the config.

![Image](https://user-images.githubusercontent.com/34404855/190352482-419a02d1-6ffb-4ba3-85c9-7123cf384e34.png)

PR #251, opened 2022-09-15
Project server in side view

### LINKED ISSUE #207: Allow users to start up dev server
The following options should be available:

- Start server (`npm start`) - enabled when server is _not running_
- Stop server (exit process) - enabled when server is i
- Restart server (exit process `npm start`) - enabled when server is _running_
- Hard restart (exit process `npm run clean` and `npm start`) - enabled when server is _running_
- Clear cache (`npm run clean`) - enabled when server is _not running_

Would be nice to show tooltips on hover of the buttons to indicate what they do. In case of the hard reset it could show a small explanation under the main tooltip text saying that it will close the server clear the cache and start it back up again.

It would be nice to have the console output below the horizontal icons for any editor-developer debugging.

I made a quick XD mock of how it could all look. Don't get attached to these icons but they were the best I could find in MUI super quickly.

### Preview

Image

karczewskiPiotr:
## Description

This PR takes care of the following issues:

- #207 
- #249 
- #250 

Initially it was focused on just #207 but the other two came up along the way with some relevance to what I was working on at the time. Might have ended up a bit too big for a PR but I guess it should be fine when not counting in style files. 😓 

The server can basically do what was described in #207. It does not accept any input and provides the constrained set of actions. I think this is best for the start as it is supposed to be an easy to interact thing for editors.  Support for links is also present in case the publication would not open in the browser due to some custom uri for the server. The output is shown so that the editor can screenshot it to provide errors to the developer as we run into such issues couple times while working with Tytus. However, for most things the provided actions should take care of any issues that the editor could come across.

> 🗒️ NOTE: The code for the terminal is fairly simple. I tried to avoid overengineering it so that if we were to expand the functionalities in the future we could use [node-pty](https://github.com/microsoft/node-pty) to make it into a fully functioning terminal and move it to the bottom of the screen like in VS Code.

> ℹ️ FYI: I came across an error with `electron-seetings` once where it said that it does not have the access to the file it manages. I can not reproduce it though. I think it might have had something to do with running `rs` in the terminal when the app was running. Couldn't reproduce it on development since and never managed to break it in production either but I thought it might be worth putting in here in case anything comes up.

#### Server functionality

server

---

#### Persisted state of sideview elements

keep

## Linked issues

Closes #207

## Correct PR Checklist

- [x] I have assigned a Milestone (if applicable)


src/renderer/components/Terminal/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+const Toolbar: React.FC<Props> = ({ project }) => {
+  const [state, setState] = useState<SERVER>(SERVER.IDLE);
+  const { t } = useTranslation('translation');
+
+  const canStopProcess = state === SERVER.IDLE || state === SERVER.STOPPING;
+  const canStartServer = state !== SERVER.IDLE;
+  const canRestartServer = state !== SERVER.SERVER_RUNNING;
+  const canHardResetServer = state !== SERVER.SERVER_RUNNING;
+  const canClearCache = state !== SERVER.IDLE;
```
LostInStatic, line 39: This logic was very confusing to me until I scrolled down to see how it was used :haha:

I'd change the naming pattern to something like `isXDisabled`. Or just reverse the logic and negate it when passing down to the components - whatever you think is more intuitive.

LostInStatic (changes requested):
Good stuff, looks really neat. 

Apart from the naming changes requested: 
- Something went wrong once, and `server:close` handler was not registered. Unfortunately, I'm unable to reproduce the issue.
- I think there are some UI issues worth thinking about. However, not a priority right now - described them in #252

I don't think that just the two points above are enough reason to block the merge, unless the error mentioned creeps up again.

LostInStatic (approved):

PR #253, opened 2022-09-18
Switch branch based on logged in user

LostInStatic:
## Description

When a publication is activated, checked out branch will depend on logged in user. If user's role is developer, `main` will be checked out, if it is editor, their corresponding `editor-USERNAME` branch is checked out, Any changes left on the previous branch will be commited as a "WIP" commit. If the branch with that name is not present, it will be created off `main`.

Additionaly, some `isomorphic-git` usage was refactored into a new handler, mostly to avoid code repetition by providing defaults and error handling. Some code was not updated in this way because of its complexity - would rather avoid touching that without a planned refactor.

### DIFF
--- a/src/main/api/git/checkout.ts
+++ b/src/main/api/git/checkout.ts
@@ -1,0 +1,48 @@
+import { mainStore as store } from 'src/main';
+import { MAIN_BRANCH } from '../../../shared/constants';
+import { selectCurrentUserData } from '../../../shared/redux/slices/currentUserSlice';
+import { activePublication } from '../../../shared/redux/slices/loadPublicationsSlice';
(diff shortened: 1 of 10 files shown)


karczewskiPiotr (approved):
@LostInStatic looks good! You preemptively took care of the only question I would have had for the checkout with https://github.com/vnLab-Lodz/PubLab/pull/253/commits/93e425310a7351d14d73c3f9450e39c3656a30e8 commit. I removed a left-over log in https://github.com/vnLab-Lodz/PubLab/pull/253/commits/ac4b13700617d969dee2bffa5e3200c889a02d97 to not block this PR.

Issue #261, opened 2022-09-27
Disable collaborators management for user not being the owner


LostInStatic:
GitHub does not allow sharing permissions to edit collaborators on personal repos - adding/deleting collaborators should be simply disabled for any user that is not the owner.

Issue #234, opened 2022-08-15
Switch the branch based on the logged in user


karczewskiPiotr:
If the logged in user is an `editor` in the project then switch the branch to `editor-[username]`. Some sort of approach needs to be thought about when there are any changes in the files when a user switches accounts. Perhaps just stashing the changes against the current branch and reinstating them when user changes the account again. It's a rare case I suppose but something to keep in mind.

karczewskiPiotr:
Important to remember about things mentioned in #244

karczewskiPiotr:
Main part of the issue tackled in #253 

> Important to remember about things mentioned in #244

Still remain to be taken care of

PR #263, opened 2022-09-28
Collaborator list handling in respect to the current user

### LINKED ISSUE #262: Allow changing the current user's role.
Current user is automatically added as a developer when creating a new project without any way to change it (as removal is disabled). Current user's role should be switchable in project settings view and in the new project dialog.

LostInStatic:
## Description

Collaborator list inputs will be disabled when the current user is not the owner. Owner is tracked in `publab-config.json` for now (it will need to be set manually for existing publications, but does not cause errors when not set).  

It is now possible to change the current user's role in project setting and while creating a new project.

## Preview

image

## Linked issues
Closes #262, #261, #234


karczewskiPiotr (changes requested):
@LostInStatic I think we can get rid of the bottom border for the select inside the table and maybe fill the whole table cell with the select. 

I think it would probably also make sense to do the following in some later PR:

- #264

LostInStatic:
UI got cleaned up, apart from the above mentioned I made the fonts match.

karczewskiPiotr (approved):

Issue #236, opened 2022-08-15
Add publish changes to the commits view


karczewskiPiotr:
Since each editor will be working on a separate branch their default commits will be made to those branches. Therefore, there is a need to some sort of `Publish changes` button in the commits view. A tooltip explaining what it does would also probably be good to have as well a confirmation for the action. 

Before the editor branch is merged into `main` a check should be performed to see if the editor is behind the `main` branch. In case it is the `main` should be first merged into the editor branch. Only after that can the editor branch be merged into `main`. This is so that any conflicts are handled on editor branches instead of `main`.

PR #265, opened 2022-10-03
Sync repo and publish changes

### LINKED ISSUE #235: Sync the repository and notify about available changes
The only branch that will need to be kept up to date for editors is the `main` branch against which merges will be done. If there are new commits in the main branch the user should be notified about it and allowed to merge them into the editor branch. I suppose a stash would need to take place behind the back to safely merge the new changes and allow the editor to continue working.

For now we can probably get away with syncing the repo when user focuses on the app window or enters the commits view. An additional safe check will be performed before publishing the editor changes to main (#236).

LostInStatic:
## Description

Sync should work like described in linked issues. If new commits are present on the main branch, the user will get a notification. Notification will not be re-displayed unless the number of commits changes. Branch comparison is rerun: 
- every time the user refocuses the window
- when the changes view is loaded 
- when active publication is changed
- when changes are published/synced

## Preview

electron_dodK1VcJPN

## Linked issues

Closes #235, #236


karczewskiPiotr:
@LostInStatic I have pulled this to test out the merging and it still left that uncommitted changes thing. I have played around with it and this finally gave me a proper merge

```ts
async merge(
      author: string,
      branchToMerge: string,
      targetBranch: string = 'editor-karczewskiPiotr'
    ) {
      await this.checkout(`remotes/origin/${targetBranch}`);
      await git.merge({
        fs,
        dir: publication.dirPath,
        ours: targetBranch,
        theirs: branchToMerge,
        author: { name: author },
      });
      await this.checkout(targetBranch);
      const data = selectCurrentUser(store.getState());
      const token = data.auth.accessToken?.value;
      await this.push({
        authToken: token!,
        remoteRef: targetBranch,
        onAuthFailure: console.error,
      });
    },
```

The uncommited changes do not show up and the branch is also pushed to the remote at the end. Only other thing that needs adjusting in the UI is the `compareBranch` should always check with remote `main` instead of local `main`.

Could also experiment with both targets as remotes to see if the checkouts can be omitted I guess but at least this works as expected even though seems weird.

LostInStatic:
@karczewskiPiotr I've incorporated you solution and tinkered with the logic a bit - seems to work smoothly now on my end. There is a bit of back-and-forth merging when publishing - merge commit is still treated like a commit, so shows up in a difference. It is still possible to be left with a hanging merge commit on the user branch side, but I think it is allright.

karczewskiPiotr (approved):
@LostInStatic I have implemented a few fixes to the merging/syncing/publishing process and added a few minor features/fixes to app itself. Here is a little video for the merging and publishing -> https://filmschool-my.sharepoint.com/:v:/g/personal/p_karczewski_vnlab_org/EQAUmxQVfz9EkKOB6FxlRs4BEBYxGkzsgQ_AdMNvZl7kaA?e=2lKBTU

You can take a look at the code and do some testing if you want. I will merge this PR in the evening to build the app after that.

PR #270, opened 2022-11-14
182 project list sort filter

### LINKED ISSUE #182: Implement sorting and filtering of the project list
The list should be sortable by headers. I suppose we should add the up and down arrows to them to signify the sorting order. The status should be used for filtering with a select.

LostInStatic:
## Description

Implemented as described in #182. `ProjectsTable.tsx` was getting a bit logic-heavy, so it was refactored into separate hooks. 

## Preview

NVIDIA_Share_0s1aye3LK7

## Linked issues

Closes #182

### DIFF
--- a/src/renderer/views/ProjectsList/subcomponents/ProjectsTable/TableHeaders.tsx
+++ b/src/renderer/views/ProjectsList/subcomponents/ProjectsTable/TableHeaders.tsx
@@ -34,10 +34,20 @@ const TableHeader = ({
     <TableRow>
       <TableCell />
       {(Object.values(SORTABLE_FIELD) as SORTABLE_FIELD[]).map((field) => (
-        <TableCell key={field}>
+        <TableCell
+          key={field}
+          sx={{
+            paddingLeft: '1rem',
+            textAlign: field === 'title' ? 'left' : 'center',
+          }}
+        >
           <Button
             variant='text'
-            sx={{ padding: 0, minWidth: 0, textAlign: 'left' }}
+            sx={{
+              padding: 0,
+              minWidth: 0,
+              textAlign: field === 'title' ? 'left' : 'center',
+            }}
             onClick={() =>
               setSortParams({
                 field,
@@ -61,7 +71,12 @@ const TableHeader = ({
           </Button>
         </TableCell>
       ))}
-      <TableCell>
+      <TableCell
+        sx={{
+          paddingLeft: '1rem',
+          textAlign: 'center',
+        }}
+      >
         <HeaderFilterSelect
           placeholder={t(`publication.status`).toLocaleUpperCase()}
           value={statusFilterValue || undefined}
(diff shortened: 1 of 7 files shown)


src/renderer/views/ProjectsList/subcomponents/ProjectsTable/TableHeaders.tsx
```diff
@@ -1,34 +1,106 @@
...
+          >
+            <Typography variant='caption' component='p'>
+              {t(`publication.${field}`)}
+              {sortParams.field === field ? (
+                orderSymbols[sortParams.direction]
+              ) : (
+                <span style={{ visibility: 'hidden' }}>
+                  {orderSymbols.ascending}
+                </span>
+              )}
```
karczewskiPiotr, line 69: @LostInStatic just wanted to ask for a bit of clarification as to why there is the need to render the invisible `span` in such case.

src/renderer/views/ProjectsList/subcomponents/ProjectsTable/hooks/useProcessedProjectList.ts
```diff
@@ -0,0 +1,41 @@
...
+  );
+
+  useEffect(() => {
+    const list = statusFilterValue
+      ? publications.filter((project) => project.status === statusFilterValue)
+      : [...publications];
+    list.sort(compareFunctions[sortParams.field]);
+    if (sortParams.direction === 'descending') list.reverse();
+    setProjectList(list);
+  }, [publications, sortParams, statusFilterValue]);
```
karczewskiPiotr, line 32: @LostInStatic I think it might make more sense here to use `useMemo` for the `projectsList` instead of `useState` + `useHook` combination since it's more of a derived value and the `setProjectsList` does not go outside the scope of this hook.

karczewskiPiotr (changes requested):
I'd only change the method for the `projectsList` derivation. Everything else seems great!

src/renderer/views/ProjectsList/subcomponents/ProjectsTable/hooks/useProcessedProjectList.ts
LostInStatic: Oh yes, definitely!

src/renderer/views/ProjectsList/subcomponents/ProjectsTable/TableHeaders.tsx
karczewskiPiotr: @LostInStatic if you want ot keep the centered things dead center you can maybe add another span on the other side of the element and that should do the trick.

LostInStatic: @karczewskiPiotr I went with a flexbox solution as discussed, the final effect IMO looks allright:


electron_7X2jfNyfHH

karczewskiPiotr: @LostInStatic that looks good 👀

karczewskiPiotr (approved):

PR #271, opened 2022-11-22
Implement context menu

### LINKED ISSUE #267: Add button to open file explorer
It would be useful to have a button in the `Files` view that allows to open the publication folder in file explorer. There are couple possibilities for this:

- top right corner next to refresh (opens `publication` folder)
- next to the go back dots in explorer header (opens current folder)
- on context menu for explorer items (opens folder or the parent folder if element is a file)
 
I suppose the context would be the most versatile one as it would also work in the commits view. I think we can go for that one.

LostInStatic:
## Description

Context menu was added, the user is able to bring it up by right-clicking. Right now, it serves two functions: 
 - Providing basic functions anywhere in the app. Basic copy and paste is implemented - copy is always available (I tried to make it visible only if some text was selected, but it was pretty buggy). Paste is visible only if an input element is active.
 - Providing access to actions defined in another component. The best way to do it is through a `useContextMenu` hook - for now it takes a list of objects describing additional menu items and returns a function that should be attached to `onContextMenu` property of a relevant element. There is an alternative way, through `ContextMenuTarget` component - it is a bit redundant, but I already had it written when I decided to use the hook instead, so left it just in case.

Context action for viewing folders in the default file manager was added to the `FileDisplay` component.

## Preview

image
image
image

## Linked issues

Closes #267


src/renderer/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,114 @@
...
+  children: React.ReactNode[];
+}
+
+type HandlerFunctionCreator = (
+  data: ContextData
+) => (event: React.MouseEvent) => void;
+
+export const MenuContext = createContext<HandlerFunctionCreator>(
+  () => () => {}
+);
```
karczewskiPiotr, line 18: I would usually make context with `undefined` as a possibility to not create these fake values. Not the biggest inconvenience here but can sometimes be a bother when making context with bigger values.

```suggestion
export const MenuContext = createContext<HandlerFunctionCreator  | undefined>(
  undefined
);
```

src/renderer/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,114 @@
...
+              data: data || {},
+              element: document.activeElement || event.currentTarget,
+              mouseX: event.clientX + 2,
+              mouseY: event.clientY - 6,
+            }
+          : null
+      );
+    },
+    []
+  );
```
karczewskiPiotr, line 45: To memo a function it's best to use the `useCallback` instead of `useMemo`.

```suggestion
  const createContextMenuHandler = useCallback(
    (data?: ContextData) => (event: React.MouseEvent) => {
      event.preventDefault();
      event.stopPropagation();
      setContextMenu(
        contextMenu === null
          ? {
              data: data || {},
              element: document.activeElement || event.currentTarget,
              mouseX: event.clientX + 2,
              mouseY: event.clientY - 6,
            }
          : null
      );
    },
    []
  );
```

src/renderer/components/ContextMenu/useContextMenu.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext, useMemo } from 'react';
+import { MenuContext } from './ContextMenu';
+import { ContextData } from './types';
+
+export default function useContextMenu(data: ContextData) {
+  const handlerCreator = useContext(MenuContext);
```
karczewskiPiotr, line 6: If we make the context have the default value of `undefined` we can then do this.

```suggestion
  const handlerCreator = useContext(MenuContext);

  if (!handlerCreator) {
    throw new Error(
      'useContextMenu can only be used within MenuContext.Provider'
    );
  }
```

It then makes it easier to figure out why the context menu is not working as an explicit error is thrown instead of an empty void function firing in silence behind the scenes.

src/renderer/components/ContextMenu/ContextMenu.tsx
karczewskiPiotr: This works well for catching missing context providers in conjuction with the comment for `src/renderer/components/ContextMenu/useContextMenu.ts`.

karczewskiPiotr (changes requested):

karczewskiPiotr (approved):

PR #272, opened 2022-11-22
Make subview hideable and improve tooltips positioning

### LINKED ISSUE #252: Improve server controls & tooltips UI
## Description

This issue serves mainly as a reminder to brainstorm solutions (or decide not to) to following issues:
 - Terminal window in sideview might be too small for some cases. Additionally, if the user does not want it to be visible, having an option to control a server without clicking 'keep visible' button would be convenient.
 - Installing dependencies can take a long time without any message in the terminal. This might be confusing, especially for a first-time user.
 - A button to manually open the website in the browser would be useful, if the user closes the browser tab by accident.
 - Immediate tooltips can be annoying if the controls are in vertical arrangement. Maybe a (configurable?) delay should be implemented?

LostInStatic:
## Description

As discussed in #252. 

## Preview

electron_3XNzIWKqkb

## Linked issues

Closes #252

### DIFF
--- a/src/renderer/components/SubviewPanel/style.ts
+++ b/src/renderer/components/SubviewPanel/style.ts
@@ -12,7 +12,7 @@ export const Panel = styled(Box)<{ isExpanded?: boolean }>(
       display: 'flex',
       minWidth: isExpanded ? width : `${buttonWidth}rem`,
       maxWidth: isExpanded ? width : `${buttonWidth}rem`,
-      background: theme.palette.primary.main,
+      background: theme.palette.background.default,
       transition: 'all 0.3s',
     };
   }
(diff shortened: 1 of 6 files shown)


src/renderer/components/SubviewPanel/style.ts
```diff
@@ -0,0 +1,40 @@
...
+    background: 'rgba(255, 255, 255, 0.1)',
+  },
+}));
+
+export const SubviewBox = styled(Box)<{ isExpanded?: boolean }>(
+  ({ isExpanded }) => ({
+    width: isExpanded ? `${subviewWidth}rem` : 0,
+    transition: 'all 0.3s',
+  })
+);
```
karczewskiPiotr, line 40: @LostInStatic I think it might better to not change the width of the `SubviewBox` but only the `Panel` one and give it hidden overflow. This way we will get rid of the weird looking text wrapping as the sidebar collapses.

karczewskiPiotr (changes requested):
We should probably make sure that the collapse button matches the color of the sidebar itself so that it works for the info sidebar in project list view. ↓

image

LostInStatic:
@karczewskiPiotr Implemented the changes, should be good now.

karczewskiPiotr (approved):
Just fixed some left over prop passing.

